PR #1, opened 2015-10-14
GeoIP Lookups from Lua

neilcook:
Added GeoIP Lookups into weakforced. Only looks up countries at the moment, not cities.

Depends on libgeoip-dev package, which I manually added as a link step to Makefile.am, as I don't know how to use autoconf yet.

Add two new lua functions:
initGeoIPDB() - this loads the ipv4 and ipv6 geoip databases. This is a pre-requisite for running the lookup function(s).
lookupCountry(comboAddress) - Returns the two-letter country code where the IP address is located. Automatically figures out which DB to use (v4 or v6). If for example the v6 DB is not available, ipv6 lookups won't work.

New GeoIP DB class defined in wforce-geoip.x, it's really simple.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -1,5 +1,7 @@
 AM_CPPFLAGS = -I pdns $(LUA_CFLAGS) $(YAHTTP_CFLAGS) -O3 -Wall -pthread 
 
+GEOIP_LIBS=-lGeoIP
+
 SUBDIRS=pdns/ext/yahttp
 
 BUILT_SOURCES=htmlfiles.h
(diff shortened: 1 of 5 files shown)


neilcook:
Cool, that was painless, thanks Bert!

ahupowerdns:
What would be even cooler if you could make it benefit from PKGCONF for
libgeoip, like this in configure.ac:

PKG_PROG_PKG_CONFIG
PKG_CHECK_MODULES([GEOIP], [geoip],[],
    AC_MSG_ERROR([Could not find libGeoIP])
)

This would automate the detection. You could even make GEOIP optional if you
wanted by removing the ERROR and rely on the HAVE_GEOIP define.

	Bert

neilcook:
Sure, I will look at this, 

Neil

PR #3, opened 2015-11-02
Time-Window based stats DB

neilcook:
This adds statistics functionality to Lua for tracking behavior of IPs/users over time.
The basic idea is that you define the set of things you want to track statistics for, using three (currently) types of stats counters: simple integer counters, hyperloglog for tracking cardinality, and countmin for tracking multisets. You then define one or more DBs to track these statistics over whatever time windows you like, using arbitrary strings or ComboAddresses as keys into the DB.

There is currently no expiration of entries in the DB, and it is entirely memory-based. I've also done no performance testing.

Really this is a first cut at looking how/if we could extend weakforce by moving a lot of the stats tracking logic from C++ into Lua, using efficient algorithms like hll and countmin to provide a lot of power (and hopefully eventually speed/efficiency). Thus making it possible for admins and implementors to write extremely powerful Lua-based anti-abuse scripts without having to ask for devs to write new functionality in C++.

There is deliberately no concept of saving data long-term. I do think this is something that is necessary (e.g. doing anomaly detection based on realtime behavior not matching long-term behavior), but could feed off the data stored in memory, e.g. by logging expired time windows. Alternatively some kind of lambda architecture where the same events are fed to a batch-processing system. Anyway, probably getting ahead of myself here...

TODO:
- Circular buffers probably not a good idea
- Need to expire old entries in a background thread or similar
- Support netmasks on ComboAddress keys so IP stats can be aggregated, not just /32 or /128
- Other types of statistics algorithms, e.g. bloomfilter for single-value sets
- Performance? Memory usage? etc.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -17,6 +17,9 @@ wforce_SOURCES = \
 	wforce-lua.cc \
 	wforce-web.cc \
 	wforce-geoip.cc \
+	twmap.cc \
+	murmur3.cc \
+	count_min_sketch.cpp \
 	dolog.hh \
 	iputils.cc iputils.hh \
 	misc.cc misc.hh \
--- a/Makefile.am
+++ b/Makefile.am
@@ -18,8 +18,6 @@ wforce_SOURCES = \
 	wforce-web.cc \
 	wforce-geoip.cc \
 	twmap.cc \
-	murmur3.cc \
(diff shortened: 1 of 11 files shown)


neilcook:
Should also add that I added wforce.conf.example, where I attempt to document how to use the new functions from Lua

PR #10, opened 2015-11-17
New Lua reporting function and initial regression tests for time windows

neilcook:
New lua report function that allows custom stats to be added from Lua.

Regression tests for exercising the timewindow functions. Makes use of the wforce console to push test-specific changes to the Lua state machine.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,6 +12,7 @@ addons:
       - liblua5.1-0-dev
       - python-virtualenv
       - libgeoip-dev
+      - geoip-database
     sources:
       - ubuntu-toolchain-r-test
 before_script:
@@ -20,11 +21,6 @@ before_script:
  - dpkg-deb -x libsodium-dev_1.0.3-1_amd64.deb ${HOME}/libsodium
  - rm libsodium13_1.0.3-1_amd64.deb libsodium-dev_1.0.3-1_amd64.deb
  - sed -i -e "s@/usr@${HOME}/libsodium/usr@" ${HOME}/libsodium/usr/lib/x86_64-linux-gnu/pkgconfig/libsodium.pc
- - wget http://geolite.maxmind.com/download/geoip/database/GeoLiteCountry/GeoIP.dat.gz http://geolite.maxmind.com/download/geoip/database/GeoIPv6.dat.gz
- - gunzip GeoIP.dat.gz
- - gunzip GeoIPv6.dat.gz
- - mkdir -p /usr/share/GeoIP
- - cp GeoIP.dat GeoIPv6.dat /usr/share/GeoIP
  - export PKG_CONFIG_PATH=${HOME}/libsodium/usr/lib/x86_64-linux-gnu/pkgconfig
  - export LD_LIBRARY_PATH=${HOME}/libsodium/usr/lib/x86_64-linux-gnu
 script:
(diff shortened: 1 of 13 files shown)


Habbie:
```
wforce-geoip.cc: In member function ‘const string WFGeoIPDB::lookupCountry(const ComboAddress&)’:
wforce-geoip.cc:31:5: error: ‘GeoIPLookup’ was not declared in this scope
     GeoIPLookup gl;
```

I haven't figured out all details but it is clear to me that this struct is in fact completely absent from the version of libgeoip that travis (and thus ubuntu 12.04) gives us. It looks like weakforced simply expects a newer version of geoip than 12.04 provides.

Habbie:
I notice, relatedly, that the pdns travis gets the geoip libs from trusty instead of precise. Sadly we don't have that kind of control on the docker-based container weakforced's travis runs in.

Options:
1. change our choice of travis env
2. first bite the 'move into pdns-tree' bullet where this is a fixed problem

Meanwhile if I can get your tests to run locally (which I should have time for on Thu) I'm fine with merging.

Habbie:
Local testing on my Mac, one failure:
```
AIL: test_invalidPasswords (test_GeoIP.TestGeoIP)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/peter/projects/powerdns/weakforced/regression-tests/test_GeoIP.py", line 16, in test_invalidPasswords
    self.assertEquals(j['status'], -1)
AssertionError: 0 != -1
```

I don't think I have geoip data installed; will check that next. I notice that there are long delays between the tests and that the wforce console is attached to my terminal - not sure it was like that for my initial small set of tests.

Habbie:
Having geoip lite data installed did not change that one test result. I did not dig in to find the actual failure reason.

Habbie:
A testrun is 65 seconds for me, as a data point.

neilcook:
Yeah, the test run is supposed to take a while, because there are a bunch of sleeps in there. These are to check that the time windows are expiring correctly, (they are set to 15 seconds, so the final test in each stage is to wait 15 seconds and then check that allow return 0 not -1).

Neil

neilcook:
It relies on a specific IP being detected as being in Japan. I’m using the DB downloaded from maxmind; if the one that ships with Ubuntu is different, then it’s possible that it thinks that the IP address is not in Japan. Why Japan? Because I heard “Big in Japan” in the office yesterday, was that Bert playing it? :)

Neil

Habbie:
Yes, that was Bert. Which IP are we talking about?

Habbie:
```
$ geoiplookup 112.78.112.20
GeoIP Country Edition: JP, Japan
GeoIP City Edition, Rev 1: JP, 32, Osaka, Osaka, 540-0008, 34.686401, 135.520004, 0, 0
```

neilcook:
112.78.112.20

dig -x 112.78.112.20 gives me 	www1710.sakura.ne.jp.

Neil

neilcook:
Hmm, not sure why it's failing then. 

If you could add the following code to a sample wforce.conf we could see what Lua thinks the country is:

-- in the main code body
initGeoIPDB() 

-- At the start of the allow function
cur_ct = lookupCountry(lt.remote) 
print(cur_ct)

And then do something like:
curl -X POST --data '{"login":"baddie", "remote": "112.78.112.20", "pwhash":"1234'$a'", "success":"false"}'     http://127.0.0.1:8084/?command=allow -u wforce:super

neilcook:
Ok I just compiled and ran on Macos, and all the tests worked. My suspicion is that your Macos environment only has an IPv4 GeoIP DB, and not v6. The GeoIP code expects both otherwise it causes an exception at initialisation time.

neilcook:
For the IPv6 DB:
pushd /usr/local/var/GeoIP/; wget -N http://geolite.maxmind.com/download/geoip/database/GeoIPv6.dat.gz; gunzip GeoIPv6.dat.gz; popd

PR #35, opened 2016-03-23
Do not install yahttp

cmouse:


cmouse:
Did you decide to do the change after all?

cmouse:
interesting, because this worked for me just fine when I made a complete reconfigure

cmouse:
you should not link yahttp dynamically anyways

neilcook:
Ok my bad, a full reconfigure and make clean fixed it.

PR #87, opened 2016-09-14
Trivial typo fixes

justinclift:


justinclift:
There's also one other potential trivial typo, though I'm unsure about it.  Just before the "API Calls" heading, there is a sentence containing:

    secret, the login, a 0 bytes and the

The "a 0 bytes" refers to the '\x00' in the code fragment above.  If the '\x00' is actually one byte (eg '\0' in C), that sentence fragment should be "a 0 byte".

And yeah, this is being pedantic. :wink:

neilcook:
> The "a 0 bytes" refers to the '\x00' in the code fragment above.

Pedantic maybe, but correct. Feel free to fix in this PR.

justinclift:
No worries at all.  Done. :smile:

PR #112, opened 2017-02-20
Support for device parsing

neilcook:
This PR adds support for parsing of device_id fields.

If "device_id" field is present in allow or report command, but no "device_attrs" field then device_attrs will be populated. The device_id field is parsed according to the "protocol" field, i.e. if http(s) then it is parsed as if it were a User-Agent string from HTTP headers, if imap(s) then it is parsed as if it were an IMAP ClientID string. Other protocols will not be parsed, as they have no string representing the device (e.g. pop, smtp).

HTTP User-Agent parsing is done with the uap-cpp library, which is copied from https://github.com/ua-parser/uap-cpp
IMAP ClientID parsing is done with hardcoded regexes for now.

New -R option to wforce is added to specify where to find the regexes needed by uap-cpp. These are retrieved from https://github.com/ua-parser/uap-core/blob/master/regexes.yaml at build time if they don't exist locally, so they will always reflect the latest version.

This PR also includes a small performance optimisation to wforce-web.cc, when is to not reallocate the pollfd array every loop, but to reallocate only if it needs to grow.


device_parser.cc
```diff
@@ -47,7 +47,6 @@ IMAPClientID IMAPClientIDParser::parse(const std::string& clientid_str) const
   boost::smatch what;
   IMAPClientID ic;
   
-  std::cout << "parsing imap string: [" << clientid_str << "]\n";
```
cmouse: unrelated change.

docs/manpages/wforce.1.md
```diff
@@ -57,8 +60,6 @@ ensure that any firewalls forward UDP on the configured ports.
 -h,--help
 :    Display a helpful message and exit.
 
--l,--local *ADDRESS*
```
cmouse: unrelated change.

regression-tests/test_Basics.py
```diff
@@ -27,7 +27,15 @@ def test_customFunc(self):
         r = self.customFunc("custom1")
         j = r.json()
         self.assertEquals(j['r_attrs']['login'], 'custom1')
-        
+
+    def test_deviceParsing(self):
```
cmouse: the regression bit should be it's own commit.

device_parser.cc
```diff
@@ -49,7 +49,6 @@ IMAPClientID IMAPClientIDParser::parse(const std::string& clientid_str) const
   
   std::cout << "parsing imap string: [" << clientid_str << "]\n";
   if (boost::regex_search(clientid_str, what, name_expr)) {
-    std::cout << "found name";
```
cmouse: unrelated change.

regression-tests/wforce-tw.conf
```diff
@@ -58,7 +58,7 @@ function twallow(lt)
              then
                 return -1, "accountStatus blocked", "accountStatus blocked", {}
              end
-        end
+	end
```
cmouse: unrelated change

wforce-lua.cc
```diff
@@ -546,6 +546,9 @@ vector<std::function<void(void)>> setupLua(bool client, bool allow_report, LuaCo
   c_lua.registerMember("policy_reject", &LoginTuple::policy_reject);
   c_lua.registerMember("attrs", &LoginTuple::attrs);
   c_lua.registerMember("attrs_mv", &LoginTuple::attrs_mv);
+  c_lua.registerMember("protocol", &LoginTuple::protocol);
```
cmouse: these should be their own commit

Makefile.am
```diff
@@ -36,6 +36,7 @@ wforce_SOURCES = \
 	sholder.hh \
 	sodcrypto.cc sodcrypto.hh \
 	luastate.hh \
+	device_parser.hh device_parser.cc \
```
cmouse: this should be in previous commit

configure.ac
```diff
@@ -13,15 +13,24 @@ PKG_CHECK_MODULES([libsodium], [libsodium], [AC_DEFINE([HAVE_LIBSODIUM], [1], [D
 PKG_CHECK_MODULES([LIBSYSTEMD], [libsystemd], [AC_DEFINE([HAVE_LIBSYSTEMD], [1], [Define to 1 if you have libsystemd-dev])], [true])
 PKG_CHECK_MODULES([LIBHIREDIS], [libhiredis], [AC_DEFINE([HAVE_LIBHIREDIS], [1], [Define to 1 if you have libhiredis-dev])], [true])
 PKG_CHECK_MODULES([LIBHIREDIS], [hiredis], [AC_DEFINE([HAVE_LIBHIREDIS], [1], [Define to 1 if you have hiredis])], [true])
+PDNS_CHECK_YAMLCPP([AC_DEFINE([HAVE_YAMLCPP], [1], [Define to 1 if you have yaml-cpp])], [true])
+AS_IF([test "x$YAMLCPP_LIBS" == "x"], [
```
cmouse: unrelated change

configure.ac
```diff
@@ -13,15 +13,24 @@ PKG_CHECK_MODULES([libsodium], [libsodium], [AC_DEFINE([HAVE_LIBSODIUM], [1], [D
...
 BOOST_FOREACH
 AC_CHECK_PROG(PERL, perl, perl)
+AC_CHECK_PROG(HAVE_WGET, wget, wget)
+AS_IF([test "x$HAVE_WGET" == "x"], [
+  AC_MSG_ERROR([wget not found, wget needed for downloading regexes.yaml and is not optional])
+])
 AX_PROG_PERL_MODULES( Swagger2::Markdown, , AC_MSG_WARN(Need to install Perl Swagger2::Markdown))
 PDNS_WITH_PROTOBUF
-AS_IF([test "x$PROTOBUF_LIBS" == "x" -o x"$PROTOC" == "x"], [
+AS_IF([test "x$PROTOBUF_LIBS" == "x" -o "x$PROTOC" == "x"], [
```
cmouse: unrelated change

configure.ac
```diff
@@ -13,15 +13,24 @@ PKG_CHECK_MODULES([libsodium], [libsodium], [AC_DEFINE([HAVE_LIBSODIUM], [1], [D
...
 AM_CONDITIONAL([LIBSYSTEMD],[test "$HAVE_LIBSYSTEMD" = "1"])
 AC_PROG_LIBTOOL
 BOOST_REQUIRE([1.42])
 BOOST_DATE_TIME
+BOOST_REGEX
 BOOST_FOREACH
 AC_CHECK_PROG(PERL, perl, perl)
+AC_CHECK_PROG(HAVE_WGET, wget, wget)
+AS_IF([test "x$HAVE_WGET" == "x"], [
+  AC_MSG_ERROR([wget not found, wget needed for downloading regexes.yaml and is not optional])
```
cmouse: unrelated change

m4/pdns_check_yamlcpp.m4
```diff
@@ -0,0 +1,121 @@
+# SYNOPSIS
+#
```
cmouse: unrelated change

wforce-web.cc
```diff
@@ -920,15 +920,23 @@ unsigned int g_num_worker_threads = WFORCE_NUM_WORKER_THREADS;
...
     // parse the array of sockets and create a pollfd array
-    struct pollfd* fds;
     int num_fds=0;
     {
       std::lock_guard<std::mutex> lock(sock_vec_mutx);
       num_fds = sock_vec.size();
-      fds = new struct pollfd [num_fds];
+      // Only allocate a new pollfd array if it needs to be bigger
+      if (num_fds > max_fd_size) {
+	if (fds)
```
cmouse: use fds != NULL

Makefile.am
neilcook: You mean I should merge with the previous commit?

configure.ac
neilcook: This is not unrelated - it's a dependency

configure.ac
neilcook: Not unrelated - it's a dependency

m4/pdns_check_yamlcpp.m4
neilcook: Not unrelated - yaml-cpp is a dependency

Makefile.am
cmouse: yes

configure.ac
cmouse: it does not match with commit message. you could move deps to their own commit

m4/pdns_check_yamlcpp.m4
cmouse: again, this could be a separate commit or make one commit about yaml-cpp

PR #116, opened 2017-03-15
Refactor

neilcook:


configure.ac
```diff
@@ -13,7 +13,12 @@ PKG_CHECK_MODULES([GETDNS], [getdns], [AC_DEFINE([HAVE_GETDNS], [1], [Define to
 PKG_CHECK_MODULES([libsodium], [libsodium], [AC_DEFINE([HAVE_LIBSODIUM], [1], [Define to 1 if you have libsodium])])
 PKG_CHECK_MODULES([LIBSYSTEMD], [libsystemd], [AC_DEFINE([HAVE_LIBSYSTEMD], [1], [Define to 1 if you have libsystemd-dev])], [true])
 PKG_CHECK_MODULES([LIBHIREDIS], [libhiredis], [AC_DEFINE([HAVE_LIBHIREDIS], [1], [Define to 1 if you have libhiredis-dev])], [true])
-PKG_CHECK_MODULES([LIBHIREDIS], [hiredis], [AC_DEFINE([HAVE_LIBHIREDIS], [1], [Define to 1 if you have hiredis])], [true])
+AS_IF([test "x$LIBHIREDIS_LIBS" == "x"], [
+  PKG_CHECK_MODULES([LIBHIREDIS], [hiredis], [AC_DEFINE([HAVE_LIBHIREDIS], [1], [Define to 1 if you have hiredis])])
+])
+AS_IF([test "x$LIBHIREDIS_LIBS" == "x"], [
+  AC_MSG_ERROR([libhiredis not found, libhiredis support is not optional])
```
cmouse: this could be "libhiredis is required", as it could be elsewhere.

wforce_exception.hh
```diff
@@ -28,3 +28,6 @@ class WforceException: public std::runtime_error
 public:
   WforceException(const std::string &a) : std::runtime_error(a) {}
 };
+
+// This means we can use PDNS classes that emit PDNSException
+typedef WforceException PDNSException;
```
cmouse: Why not just rename, unless you are actually using powerdns sources directly?

configure.ac
neilcook: Yes good point, should make it standard

wforce_exception.hh
neilcook: Well there are a couple of places where I use pdns sources directly, so this is "insurance". It makes merging changes a lot easier as well.

PR #117, opened 2017-03-16
Add docker config for elasticsearch and logstash

neilcook:
Elasticsearch is the chosen DB to store wforce reports long-term.
Reports arrive via Logstash, using the UDP input. Logstash acts
as a NamedReportSink for wforce.
Kibana support in the Docker config is onoing due to issues with
performance on MacOS. Currently the kibana config stores the json
for a bunch of pre-configured views and dashboards for wforce data.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -90,6 +90,7 @@ noinst_HEADERS = \
 	luastate.hh \
 	device_parser.hh
 
+BUILT_SOURCES = replication.pb.cc replication.pb.h
 replication.pb.cc: replication.proto
 	$(AM_V_GEN)$(PROTOC) -I $(top_srcdir) --cpp_out=./ $<
 
--- a/Makefile.am
+++ b/Makefile.am
@@ -5,7 +5,7 @@ UAP_REGEX_FILE=regexes.yaml
 
 SUBDIRS=ext docs
 
-EXTRA_DIST= wforce.conf README.md wforce.service.in wforce.conf.example replication.proto $(UAP_REGEX_FILE)
+EXTRA_DIST= wforce.conf README.md wforce.service.in wforce.conf.example replication.proto $(UAP_REGEX_FILE) docker/Makefile.am
 
 CLEANFILES = replication.pb.cc replication.pb.h wforce.service $(UAP_REGEX_FILE)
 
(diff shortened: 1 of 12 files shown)


configure.ac
```diff
@@ -36,6 +36,10 @@ AC_CHECK_PROG(HAVE_WGET, wget, wget)
 AS_IF([test "x$HAVE_WGET" == "x"], [
   AC_MSG_ERROR([wget not found, wget needed for downloading regexes.yaml and is mandatory])
 ])
+AC_CHECK_PROG(DOCKER, docker-compose, docker-compose)
+AS_IF([test "x$HAVE_WGET" == "x"], [
+  AC_MSG_ERROR([docker-compose not found, and is mandatory])
```
cmouse: again, i would probably use word required.

Makefile.am
```diff
@@ -3,7 +3,7 @@ ACLOCAL_AMFLAGS = -I m4
 
 UAP_REGEX_FILE=regexes.yaml
 
-SUBDIRS=ext docs docker
```
cmouse: this might be a bad idea. it's probably a lot better idea to use EXTRA_ in the docker/Makefile.am for the targets, so they don't get built automatically.

neilcook: Yes agreed this is definitely a bad idea :) I've changed with the latest commit to remove it entirely, so that docker must be built explicitly. The docker stuff is mainly for regression in travis, so that seems like the best approach.

configure.ac
neilcook: ok will change mandatory to required

Makefile.am
neilcook: Ok that didn't work, I will do as you suggest :)

cmouse:
Did you test out of tree build?

Makefile.am
```diff
@@ -86,7 +86,7 @@ noinst_HEADERS = \
 	device_parser.hh
 
 replication.pb.cc: replication.proto
-	$(AM_V_GEN)$(PROTOC) --cpp_out=./ $<
+	$(AM_V_GEN)$(PROTOC) -I $(top_srcdir) --cpp_out=./ $<
```
cmouse: this requires GENERATED_SOURCES in Makefile.am for the target files.

neilcook: Added BUILT_SOURCES

PR #126, opened 2017-04-07
Enable GeoIP DBs to be reloaded

neilcook:
Previously the GeoIP DB would be loaded and new changes would be ignored.
This behaviour is very fast and thread safe, but meant that changes to the
GeoIP DB were ignored, which is non-optimal.
This commit adds the GEOIP_CACHE_CHECK flag to the DB open function, which
means that the DB will be reloaded if it changes on disk. However this change
now means that lookups are no longer thread-safe, so a mutex has been added
to protect all lookups.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -41,6 +41,8 @@ wforce_SOURCES = \
 	sholder.hh \
 	sodcrypto.cc sodcrypto.hh \
 	luastate.hh \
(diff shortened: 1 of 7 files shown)


neilcook:
Ok now this uses RW locks as per our discussion. I'll cherrypick to #127 once you review and approve

PR #127, opened 2017-04-10
Enable GeoIP DB to be reloaded upon update

neilcook:
Previously the GeoIP DB would be loaded and new changes would be ignored.
This behaviour is very fast and thread safe, but meant that changes to the
GeoIP DB were ignored, which is non-optimal.
This commit adds a new command - reloadGeoIPDBs() which will reload all the
initialised DBs. Lookups are now protected by a read lock, while the reload
function uses a write lock.
This is a v1.0.x backport of of #127

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -23,6 +23,8 @@ wforce_SOURCES = \
 	perf-stats.cc perf-stats.hh \
 	sholder.hh \
 	sodcrypto.cc sodcrypto.hh \
+	lock.hh \
+	pdnsexception.hh \
 	sstuff.hh ext/luawrapper/include/LuaContext.hpp \
 	ext/incbin/incbin.h \
 	ext/incbin/incbin.h \
@@ -57,6 +59,8 @@ noinst_HEADERS = \
 	blacklist.hh \
 	wforce-geoip.hh \
 	wforce.hh
+	lock.hh \
+	pdnsexception.hh
 
 if HAVE_SYSTEMD
 SYSTEMD_TYPE = notify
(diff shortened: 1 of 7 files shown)


cmouse:
you sure that mutex is needed?

neilcook:
Yep, mutex is needed because when you enable the GEOIP_CHECK_CACHE flag, the DB can be reloaded at anytime. See https://github.com/maxmind/geoip-api-c#threads

cmouse:
so... how is this different from #126?

cmouse:
also. i would perhaps recommend doing controlled reload for the database instead. this way you would not need to serialize all lookups and make this effectively a bottleneck.

neilcook:
It’s the same. #127 is just cherrypicking #126 to the 1.0.x branch for IOL.

neilcook:
I thought about doing a controlled reload. The problem is that the functions are opaque, you don’t have a way to find out if the DB needs reloading. I *could* switch to loading the files directly i.e. by specifying the file names, but that gets really ugly really quickly because then you need to force the wforce admin to know where the files are located and what they are called and configure them manually. Currently I let the GeoIP library handle all of that - I’d be loath to change the way that works.

The performance of GeoIP DB (lookups/sec) is ~1-2 orders of magnitude greater than the performance of wforce so I’m not too worried about this becoming a bottleneck. Lookups are extremely low latency.

One thing I could do is make GEOIP_CACHE_CHECK an optional flag, and then only use the mutex if the flag is set.

cmouse:
you could also periodically check if the file has changed... not sure how various caching semantics work here

neilcook:
> you could also periodically check if the file has changed... not sure how various caching semantics work here

Yeah but the point is that currently I don’t even know the names or locations of the DB files, I let the geoip library handle that and just refer to DBs using the GeoIPDBType enum.

neilcook:
Checkout the original PR #126 - this now contains the RW locks as per our discussion. I'll cherrypick to this PR once you review and approve.

PR #128, opened 2017-04-11
Docs: correct possible values for protocol field

neilcook:


cmouse:
not sure if dovecot has imaps in %s expansion

neilcook:
yeah I guessed not, probably not pop3s either. That's ok though - Dovecot will always send pop3 and imap. I had thought about simply removing the "s" options, and adding an extra boolean field to indicate encryption. Or maybe not caring if users are using encryption or not particularly since it's so often offloaded onto a proxy and we don't even know if they are.

cmouse:
i would recommend field indicating transport security

neilcook:
Added new tls field to indicate transport security and modify docs for this and protocol field

cmouse:
i wonder if secure would be better name for it...

neilcook:
I thought about that but I thought it wasn't specific enough - it's a bit ambiguous.

cmouse:
its just that if you restrict it to tls you ignore various other things, like intranet or localhost or ipsec (yes, rare). also, does tls imply trusted? we can go with tls if you want.

neilcook:
> its just that if you restrict it to tls you ignore various other things, like intranet or localhost or ipsec (yes, rare).

Yeah, the thing is I don't think I want to know all those other things, or at least they're not really actionable. I think I'l just stick with tls. I don't think tls implies anything other than that the connection used transport security

PR #129, opened 2017-04-19
Webhook changes to support logstash HTTP input

neilcook:
We'd like to support webhooks as a mechanism to send data to logstash in addition using a reliable protocol, in addition to the unreliable UDP report sink mechanism.
- This PR adds support for that, after testing with logstash revealed that it doesn't support chunked HTTP.
- Also add configuration parameter to control the number of outbound HTTP connections wforce makes to the downstream HTTP server for each web hook.
- Use webhooks instead of UDP in the gendata scripts
- Also a commit to document the reloadGeoIPDBs() command
- Add an elasticsearch template (managed by logstash) to ensure that data is always consistent in ES, even when arriving via different input methoda
- Update the kibana saved objects to reflect the new ES schema

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -4,6 +4,7 @@ services:
     build: logstash
     ports:
       - "14501:4501/udp"
+      - "18080:8080/tcp"
     command: -f /etc/logstash/conf.d/
     volumes:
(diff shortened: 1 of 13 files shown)


neilcook:
@cmouse ping

PR #144, opened 2017-06-07
Trackalert Performance and Documentation

neilcook:
- Decouple web server thread pool from report thread pool - means that slow reports don't chew up web server threads, as they now run in their own thread pool. Also means that the report API returns OK before the report function is run, which is fine.
- Initial Trackalert README.md
- Man pages for trackalert
- Swagger documentation for trackalert API


common/misc.cc
```diff
@@ -864,3 +864,12 @@ std::string getDirectoryPath(const std::string& filename)
   else
     return ".";
 }
+
+std::string getFileFromPath(const std::string& filename)
```
cmouse: http://www.boost.org/doc/libs/1_31_0/libs/filesystem/doc/path.htm

neilcook:
Cool, thx - I’ll use that instead.

common/misc.cc
```diff
@@ -858,18 +859,13 @@ unsigned int pdns_stou(const std::string& str, size_t * idx, int base)
...
 std::string getDirectoryPath(const std::string& filename)
 {
-  size_t found = filename.find_last_of("/\\");
-  if (found != string::npos)
-    return filename.substr(0, found);
-  else
-    return ".";
+  boost::filesystem::path my_path(filename);
+  boost::filesystem::path branch_path = my_path.branch_path();
+  return branch_path.empty() ? "." : branch_path.string();
```
cmouse: should it use native?

neilcook:
I thought about using native, but:

1) I have no desire for this to work on anything other than POSIX compliant systems
2) Generic works on both Linux and Mac

I’d also be concerned about side-effects with other code, e.g. assuming that g_configDir was in posix format.

Neil

PR #156, opened 2017-08-08
Blacklist

neilcook:
Refactor blacklist and add netmask support so whole IP ranges can be blacklisted

### DIFF
--- a/blacklist.hh
+++ b/blacklist.hh
@@ -23,6 +23,7 @@
 #pragma once
 #include <time.h>
 #include "misc.hh"
+#include "lock.hh"
 #include "iputils.hh"
 #include <atomic>
 #include <mutex>
@@ -122,7 +123,7 @@ private:
   blacklist_t ip_blacklist;
   blacklist_t login_blacklist;
   blacklist_t ip_login_blacklist;
-  mutable std::mutex mutx;
+  mutable pthread_rwlock_t bl_rwlock = PTHREAD_RWLOCK_INITIALIZER;
   bool persist=false;
   bool persist_replicated=false;
   std::string redis_server;
--- a/blacklist.hh
+++ b/blacklist.hh
@@ -63,22 +63,22 @@ public:
   void addEntry(const std::string& login, time_t seconds, const std::string& reason);
   void addEntry(const ComboAddress& ca, const std::string& login, time_t seconds, const std::string& reason);
 
-  bool checkEntry(const ComboAddress& ca);
-  bool checkEntry(const std::string& login);
-  bool checkEntry(const ComboAddress& ca, const std::string& login);
+  bool checkEntry(const ComboAddress& ca) const;
+  bool checkEntry(const std::string& login) const;
+  bool checkEntry(const ComboAddress& ca, const std::string& login) const;
 
-  bool getEntry(const ComboAddress& ca, BlackListEntry& ret);
-  bool getEntry(const std::string& login, BlackListEntry& ret);
-  bool getEntry(const ComboAddress& ca, const std::string& login, BlackListEntry& ret);
+  bool getEntry(const ComboAddress& ca, BlackListEntry& ret) const;
+  bool getEntry(const std::string& login, BlackListEntry& ret) const;
+  bool getEntry(const ComboAddress& ca, const std::string& login, BlackListEntry& ret) const;
 
   void deleteEntry(const Netmask& nm);
   void deleteEntry(const ComboAddress& ca);
   void deleteEntry(const std::string& login);
   void deleteEntry(const ComboAddress& ca, const std::string& login);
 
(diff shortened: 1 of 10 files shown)


common/iputils.hh
```diff
@@ -722,13 +722,31 @@ public:
...
+	  parent->left.reset();
+	else
+	  parent->right.reset();
+	// now recurse up to the parent
+	cleanup_tree(parent);
+      }
+    }
+  }
+  
+  //<! Removes key from TreeMap. This optionally cleans up the tree.
```
cmouse: the cleanup here is slightly misleading, and I think it should probably be moved to ctor as parameter instead, since it does not *really* cleanup the whole tree. either the tree cleans itself or it doesn't. inconsistently using this can lead into rather funny tree.

common/iputils.hh
```diff
@@ -722,13 +722,31 @@ public:
...
+  void erase(const key_type& key, bool cleanup=false) {
     TreeNode *node = root.get();
 
     // no tree, no value
     if ( node == nullptr ) return;
-
+
```
cmouse: whitespace

common/iputils.hh
```diff
@@ -839,6 +859,21 @@ public:
     return match(&ip);
   }
 
+  bool lookup(const ComboAddress* ip, Netmask* nmp) const
+  {
+    const auto &ret = tree.lookup(*ip);
+    if (ret) {
+      if (nmp) *nmp = ret->first;
```
cmouse: `if nmp != null`

common/iputils.hh
```diff
@@ -855,6 +890,18 @@ public:
...
+  void deleteMask(const Netmask& nm)
+  {
+    tree.erase(nm, true);
+  }
+
+  void deleteMask(const std::string& ip)
+  {
+    if (!ip.empty())
+      deleteMask(Netmask(ip));
+  }
```
cmouse: what if IP is empty?

blacklist.hh
```diff
@@ -86,13 +89,15 @@ public:
   bool loadPersistEntries();
 
   void purgeEntries();
+
```
cmouse: whitespace

blacklist.hh
```diff
@@ -114,10 +119,11 @@ private:
 
   const char* bl_names[4] = { "ip_bl", "login_bl", "ip_login_bl", NULL };
   const char* key_names[4] = { "ip", "login", "ip_login", NULL };
+  NetmaskGroup ipbl_netmask;
   blacklist_t ip_blacklist;
   blacklist_t login_blacklist;
   blacklist_t ip_login_blacklist;
-  std::mutex mutx;
+  mutable pthread_rwlock_t bl_rwlock = PTHREAD_RWLOCK_INITIALIZER;
```
cmouse: why is mutable used here?

docs/manpages/wforce.conf.5.md
```diff
@@ -305,8 +305,8 @@ configuration or within the allow/report/reset functions:
 		local my_city = gip_record.city
 		local my_latitude = gip_record.latitude
 
-* newCA(\<IP[:port]\>) - Create and return an object representing an IP
-  address (v4 or v6) and optional port. The object is called a
+* newCA(\<IP[:port]\>) - Create and return an object representing an
+  IP address (v4 or v6) and optional port. The object is called a
```
cmouse: this change seems pointless.

docs/swagger/wforce_api.7.yml
```diff
@@ -6,8 +6,8 @@ info:
   contact:
     name: "Dovecot Oy"
   license:
-    name: "GPL2"
-    url: http://github.com/PowerDNS/weakforced/master/LICENSE
+    name: "GPL3"
```
cmouse: would be nice if license changes would be in their own pull request.

common/iputils.hh
neilcook: Yes agreed

common/iputils.hh
neilcook: then nothing will be deleted

blacklist.hh
neilcook: otherwise, the read-only functions that are declared const won't compile

docs/manpages/wforce.conf.5.md
neilcook: emacs auto-wrap silliness

common/iputils.hh
cmouse: this should probably be mentioned somewhere that it's OK.

docs/swagger/wforce_api.7.yml
neilcook: Actually it's not a license change, that was already done - this is just bringing the "name" inline with what is actually contained in LICENSE

PR #161, opened 2017-08-21
New Report API

neilcook:
This consists of:

- HTTP REST API to retrieve login and device details, as well as confirm/unconfirm logins on behalf of users

This is written as a python flask module, and includes regression tests as part of the existing regression framework.

Does not include any deployment stuff yet (e.g. uwsgi ini files, or nginx config).


report_api/app/report.py
```diff
@@ -27,6 +24,17 @@
     }
 }
 
+app.config.from_pyfile('report.cfg')
```
cmouse: is there some point in moving these around?

report_api/app/report.py
```diff
@@ -54,16 +62,16 @@ def constructQuery(j, query):
...
         response = elastic.search(index=app.config['ELASTICSEARCH_INDEX'],
                                   body=login_query,
                                   _source=True)
     except elasticsearch.TransportError as err:
-        print("Elasticsearch transport error: ", err.error)
+        app.logger.error("Elasticsearch transport error: %s remote_ip=%s", err.error, client_ip)
         return None
     except elasticsearch.ElasticsearchException:
-        print("Elasticsearch exception")
+        app.logger.error("Elasticsearch Exception error remote_ip=%s", client_ip)
```
cmouse: maybe include the extracted reason too? this is bit like "generic exception"?

report_api/app/report.py
```diff
@@ -114,41 +122,57 @@ def makeDevicesResponse(es_response):
...
         return make_response(jsonify({'error_msg': 'Invalid query parameters - must supply login'}), 400)  
     my_query = constructQuery(request.json, global_query)        
-    es_response = queryElastic(my_query)
+    es_response = queryElastic(my_query, client_ip)
     response = makeLoginsResponse(es_response)
     
     if response == None:
         return make_response(jsonify({'error_msg': 'Elasticsearch query failed'}), 500)
     else:
+        app.logger.info("/logins endpoint returning logins for login=%s remote_ip=%s", request.json['login'], client_ip)
```
cmouse: this feels like debug log

report_api/app/report.py
```diff
@@ -171,15 +195,17 @@ def confirm():
     try:
         response = elastic.update(index=index,doc_type="wforce_report",id=id, body={ 'doc': { 'user_confirmation': user_confirmation }},refresh=True)
     except elasticsearch.TransportError as err:
-        print("Elasticsearch transport error: ", err.error)
+        app.logger.error("Elasticsearch update exception (%s) trying to update doc id=%s in index=%s remote_ip=%s", err.error, id, index, client_ip)
```
cmouse: why not use these on line 66 too?

report_api/run.py
```diff
@@ -1,3 +1,3 @@
 #!flask/bin/python
-from app import app
+from wforce import app
```
cmouse: could these be part of the initial commit?

report_api/requirements.txt
```diff
@@ -0,0 +1,12 @@
+click==6.7
```
cmouse: awfully strict versions, could these be bit more liberal to allow e.g. minor updates?

report_api/runreport.py
```diff
@@ -1,3 +1,3 @@
-#!flask/bin/python
+#!.venv/bin/python
```
cmouse: this should not be necessary, as activate should put stuff into path?

report_api/runreport.py
```diff
@@ -1,3 +1,5 @@
 #!.venv/bin/python
+import sys
+sys.path.append(sys.path[0])
```
cmouse: it should already be there.

cmouse:
Also while I appreciate the atomicity of the commits, I believe you could just roll those related to the new report_api into single commit, since it's not really merged yet either.

report_api/app/report.py
neilcook: Not sure what you mean by that?

report_api/app/report.py
neilcook: yep will do

report_api/app/report.py
neilcook: agreed

report_api/run.py
neilcook: If I roll it up into a single commit then yes :)

report_api/requirements.txt
neilcook: Yeah probably - they are autogenerated currently

report_api/runreport.py
neilcook: good point.

report_api/app/report.py
cmouse: lines 9-14 are moved to lines 27-30. they aren't changed in any way.

neilcook: I needed the config line after the new logging definitions so it can use them. Also I thought that imports should logically be all together for clarity

report_api/app/report.py
neilcook: It turns out that the second exception is indeed generic exception, and that all the exceptions which include an error will be in the previous exception class TransportError.

report_api/app/report.py
neilcook: Same comment as for line 66.

report_api/runreport.py
neilcook: I added it specifically because it wasn't there, e.g. When running it from the regression directory (directly - not using run.sh because that creates another virtualenv which causes issues)

neilcook:
Ok refactored and rebased

PR #174, opened 2017-10-06
fix docker-compose check and make it optional if asked to be built

wrosenauer:
I understand that --enable-docker is an explicit option but docker-compose was checked mandatorily


wrosenauer:
oh, and in addition it was checking for the wrong result ;-)

neilcook:
Looks good - thanks!

PR #175, opened 2017-10-09
Bugs

neilcook:


docs/Makefile.am
```diff
@@ -1,5 +1,10 @@
-MANPAGES_TARGET = wforce.1 wforce.conf.5 wforce_webhook.5 trackalert.1 trackalert.conf.5
-APIDOC = wforce_api.7 report_api.7 trackalert.7
+if WITH_TRACKALERT
+TRACKALERT_MANPAGES_TARGET = trackalert.1 trackalert.conf.5
```
cmouse: you could do this with
`MANPAGES_TARGET += trackalert.1 trackalert.conf.5`
instead.

neilcook: true, although with all the other trackalert conditionals, I use the "long-way-round" method as well.

PR #179, opened 2017-10-19
Refactor webhooks

neilcook:
Previously webhooks use the ctpl library, which is a thread-pool based system. However this means that if the set of webhook threads and connections are not able to keep up with the throughput of webhooks being passed to them, then webhooks would be dropped on the floor. This is inevitable, however there was no ability to control the size of the queue, nor was there any logging to say that webhooks were being dropped because the queue was full.
This PR changes the implicit queue used in ctpl to an explicit queue. We create a set of web hook threads, which wait for new events to be added to the queue. When the producer finds the queue to be too big, it explicitly logs the fact that it's dropping the web hook on the floor.
The PR also changes from using a single easy-curl connection per thread to using a the multi-curl interface for each thread, where each multi-curl is associated with a configurable number of easy-curl connections. This reduces the need for a large number of threads to achieve good performance, as well as allowing webhooks to make use of HTTP 1.1 pipelining and HTTP 2.0 multiplexing where available.
Finally this PR adds support for the boost unit test framework, and includes a trivial unit test for the MiniCurl class. I originally wanted to include more comprehensive tests for the curl and web hook classes, but unit tests are more appropriate for standalone classes, and the tests in the regression directory already cover webhooks in detail. However now the boost unit test framework is implemented, we can start to add more unit tests in the future, including existing classes.


common/webhook.cc
```diff
@@ -28,14 +28,26 @@
...
+WebHookRunner::WebHookRunner()
 {
-  curl_global_init(CURL_GLOBAL_ALL);
 }
 
-void WebHookRunner::setNumThreads(unsigned int num_threads)
+void WebHookRunner::setNumThreads(unsigned int new_num_threads)
 {
-  p.resize(num_threads);
+  num_threads = new_num_threads;
```
cmouse: why rename the parameter? you could just use this->num_threads = num_threads.

common/webhook.cc
```diff
@@ -46,81 +58,103 @@ void WebHookRunner::setMaxConns(unsigned int max_conns)
...
-    if (auto ccs = cc.lock())
-      p.push(_runHookThread, event_name, hook, hook_data, ccs);
+    WebHookQueueItem wqi = {
+      event_name,
+      hook,
+      std::make_shared<std::string>(hook_data)
+    };
+    {
+      std::lock_guard<std::mutex> lock(queue_mutex);
+      if (queue.size() > max_queue_size) {
```
cmouse: this seems to allow max_queue_size+1 item in queue

common/webhook.cc
```diff
@@ -132,8 +166,8 @@ bool WebHookRunner::_runHook(const std::string& event_name, std::shared_ptr<cons
   mch.insert(std::make_pair("X-Wforce-HookID", std::to_string(hook->getID())));
   if (hook->hasConfigKey("secret"))
     mch.insert(std::make_pair("X-Wforce-Signature",
-			      Base64Encode(calculateHMAC(hook->getConfigKey("secret"),
-							 hook_data, HashAlgo::SHA256))));
+                              Base64Encode(calculateHMAC(hook->getConfigKey("secret"),
+                                                         hook_data, HashAlgo::SHA256))));
```
cmouse: unrelated.

common/webhook.cc
```diff
@@ -143,24 +177,7 @@ bool WebHookRunner::_runHook(const std::string& event_name, std::shared_ptr<cons
   }
   
   vdebuglog("Webhook id=%d starting for event (%s) to url (%s) with delivery id (%s) and hook_data (%s)",
-	   hook->getID(), event_name, hook->getConfigKey("url"), b64_hash_id, hook_data);
+            hook->getID(), event_name, hook->getConfigKey("url"), b64_hash_id, hook_data);
```
cmouse: unrelated change.

common/webhook.hh
```diff
@@ -437,32 +438,44 @@ struct CurlConnection {
...
 private:
-  ctpl::thread_pool p;
+  std::queue<WebHookQueueItem> queue;
+  std::mutex queue_mutex;
+  std::condition_variable cv;
+  unsigned int max_queue_size = QUEUE_SIZE;
   unsigned int max_hook_conns = MAX_HOOK_CONN;
-  std::mutex  conn_mutex;
-  CurlConnMap conns;
+  unsigned int num_threads = NUM_WEBHOOK_THREADS;
```
cmouse: this could also happen in ctor, like `WebHookRunner() : max_queue_size(QUEUE_SIZE) : ...`

configure.ac
```diff
@@ -32,6 +32,8 @@ BOOST_REGEX
 BOOST_FOREACH
 BOOST_SYSTEM
 BOOST_FILESYSTEM
+AM_CONDITIONAL([HAVE_BOOST_GE_148], [test "$boost_major_version" -ge 148])
+PDNS_ENABLE_UNIT_TESTS
```
cmouse: rename to `WFORCE`?

trackalert/trackalert-lua.cc
```diff
@@ -435,6 +435,13 @@ vector<std::function<void(void)>> setupLua(bool client, bool multi_lua,
     c_lua.writeFunction("setNumWebHookThreads", [](unsigned int num_threads) { });
   }
 
+  if (!multi_lua) {
+    c_lua.writeFunction("setWebHookQueueSize", [](unsigned int queue_size) { g_webhook_runner.setMaxQueueSize(queue_size); });
+  }
+  else {
+    c_lua.writeFunction("setWebHookQueueSize", [](unsigned int queue_size) { });
```
cmouse: maybe this could emit warning?

wforce-lua.cc
```diff
@@ -845,6 +845,21 @@ vector<std::function<void(void)>> setupLua(bool client, bool allow_report, LuaCo
     c_lua.writeFunction("setNumWebHookThreads", [](unsigned int num_threads) { });
   }
 
+  if (!allow_report) {
+    c_lua.writeFunction("setWebHookQueueSize", [](unsigned int queue_size) { g_webhook_runner.setMaxQueueSize(queue_size); });
+  }
+  else {
+    c_lua.writeFunction("setWebHookQueueSize", [](unsigned int queue_size) { });
```
cmouse: maybe this could emit warning?

wforce-lua.cc
```diff
@@ -845,6 +845,21 @@ vector<std::function<void(void)>> setupLua(bool client, bool allow_report, LuaCo
...
+  }
+
+  if (!allow_report) {
+    c_lua.writeFunction("setNumWebHookConnsPerThread", [](unsigned int num_conns) { g_webhook_runner.setMaxConns(num_conns); });
+  }
+  else {
+    c_lua.writeFunction("setNumWebHookConnsPerThread", [](unsigned int num_conns) { });
+  }
```
cmouse: maybe this could emit warning?

trackalert/trackalert-lua.cc
neilcook: Warning would be bad here, as this will intentionally be called for every lua state, but it only "makes sense" for one particular lua state (same goes for the other similar examples)

configure.ac
neilcook: Maybe but I have a bunch of these from pdns, and I update them on a regular basis so it would be a pita

common/webhook.hh
neilcook: Yes but I like them here as for me it's easier to understand when looking at the header file

common/webhook.cc
neilcook: true

common/minicurl.cc
```diff
@@ -170,3 +202,156 @@ bool MiniCurl::postURL(const std::string& url,
...
+      if (mc == CURLM_OK) {
+        auto start_time = std::chrono::steady_clock::now();
+        
+        mc = curl_multi_wait(d_mcurl, NULL, 0, 1000, &numfds);
+
+        auto wait_time_ms = std::chrono::duration_cast<std::chrono::milliseconds>(std::chrono::steady_clock::now() - start_time);
+        
+        if (mc == CURLM_OK) {
+          struct CURLMsg *m;
+          do {
```
cmouse: this could be easier to read with `while((m = curl_multi_info_read(d_mcurl, &msgq)) != NULL) {`

common/minicurl.cc
```diff
@@ -170,3 +202,156 @@ bool MiniCurl::postURL(const std::string& url,
...
+            int msgq = 0;
+            m = curl_multi_info_read(d_mcurl, &msgq);
+            if (m && (m->msg == CURLMSG_DONE)) {
```
cmouse: and you could remove the `m &&` part.

common/minicurl.cc
```diff
@@ -170,3 +202,156 @@ bool MiniCurl::postURL(const std::string& url,
...
+                }
+              }
+              mpr.id = minic->getID();
+              post_ret.emplace_back(std::move(mpr));
+            }
+          } while (m);
+          if (!numfds) {
+            // Avoid busy looping when there is nothing to do
+            // Inspired by similar code in curl_easy_perform
+            if (wait_time_ms.count() <= 10) {
```
cmouse: this code seems somehow overcly complicated. at least it could use some explanation.

common/minicurl.cc
```diff
@@ -170,3 +202,156 @@ bool MiniCurl::postURL(const std::string& url,
...
+            }
+          }
+          else {
+            /* got file descriptor, restart counter */
+            without_fds = 0;
+          }
+        }
+        else { // curl_multi_wait failed
+          errlog("curl_multi_wait failed, code %d.n", mc);
+          break;
```
cmouse: curl has no string errors?

neilcook: good point :)

common/minicurl.cc
```diff
@@ -21,18 +21,34 @@
...
 #include <curl/curl.h>
 #include <stdarg.h>
 #include <string.h>
 #include <sstream>
+#include "wforce_exception.hh"
 #include "dolog.hh"
 
+bool MiniCurl::initCurlGlobal()
+{
+  // This is guaranteed to be called only once
```
cmouse: how is it exactly guaranteed to called only once?

common/Makefile.am
```diff
@@ -49,3 +49,40 @@ noinst_HEADERS = \
...
+	@echo "Run ./configure --enable-unit-tests"
+endif
+
+testrunner_SOURCES = \
+	testrunner.cc \
+	test-minicurl.cc
+
+testrunner_LDFLAGS = \
+	$(AM_LDFLAGS) \
+	-pthread \
```
cmouse: I *think* boost already adds `-pthread` flag.

cmouse: Also i think you should look at m4/pdns_check_os.m4

common/minicurl.cc
neilcook: The comment refers to the initialisation of the static below

common/Makefile.am
neilcook: Well no harm to do it twice

cmouse: except if you are not compiling on linux =)

common/minicurl.cc
cmouse: I can easily imagine it getting called more than once. If I am wrong, you could update the comment to explain why.

common/Makefile.am
neilcook: wforce is not designed to work on any OS that doesn't have pthreads

common/minicurl.cc
neilcook: Well you're right in that the initCurlGlobal() function will get called many times. But the curl_global_init call will only happen once. I'll clarify that in the comment

common/Makefile.am
cmouse: except some os'es have extra things you need to do, and it's not always called -pthread. Some have -lpthread instead.

neilcook: Ah I see

common/minicurl.cc
cmouse: I still don't get it. =)

neilcook: I need curl_global_init() to called only once. Doing it as part of the initialisation of a static variable is a way to achieve that, since statics are only initialized once.

Issue #183, opened 2017-12-14
Add allow/report/reset etc. stats to the 5 min stats logging


neilcook:
Currently:
2017-12-14T14:37:53.571193+00:00 dovauth-ch2g-03o wforce[28622]: stats last 300 secs: WTW_0_1=20194 WTW_1_10=1 WTW_10_100=0 WTW_100_1000=0 WTW_Slow=0 WTR_0_1=20175 WTR_1_10=19 WTR_10_100=0 WTR_100_1000=0 WTR_Slow=0

Proposed - something like:
2017-12-14T14:37:53.571193+00:00 dovauth-ch2g-03o wforce[28622]: stats last 300 secs: WTW_0_1=20194 WTW_1_10=1 WTW_10_100=0 WTW_100_1000=0 WTW_Slow=0 WTR_0_1=20175 WTR_1_10=19 WTR_10_100=0 WTR_100_1000=0 WTR_Slow=0 allow=30293 report=29283 reset=2345

Need to think about whether custom endpoints should be included.

neilcook:
@sshipway wrote:
It would be very useful for the existing stats call to also provide counters for the return codes, not just for the function calls.

IE, as well as counts on number of allow() and report() calls made, also counts on the number of allow() calls split by return code 0, -1, or >0.

Even better would be to have a generic count, so in the Lua code you could use something like
addstat(key) followed by increment(key) in an allow or report function, and then the normal stats() call would also return "counter.key: 1" or similar. I know we could implement this in lua using a database, but it would be slow.

Counters like this could then be fed into your favourite monitoring package (MRTG, Zabbix, Nagio, Cacti...) to obtain graphs of weakforced performance and behaviour.

sshipway:
The new custom stats feature is awesome; great to see this.  However, can we have separate stats counters gathered based on the *return value* from the allow function (denies, tarpits, permits)?  Currently, there is no stat for the number of requests denied due to blacklisting, for example, which is a useful indicator of how well we're doing.  A custom stat in the lua couldn't do this as blacklist is checked before it gets there.

neilcook:
Well the stats() function does count “denies” currently. It just doesn’t split out blacklisted vs a Lua-based deny for example,

Neil

neilcook:
This should give you what you want:
> showCommandStats() 
addBLEntry=6
allow=6
allow_allowed=21
allow_blacklisted=6
allow_denied=5
allow_tarpitted=0
countLogins=0
[...]

PR #191, opened 2018-03-07
Improved Stats functions

neilcook:
Inspired by #183 and #187

### DIFF
--- a/common/perf-stats.cc
+++ b/common/perf-stats.cc
@@ -34,15 +34,15 @@
 TWStatsDB<unsigned int> g_perfStats("perfStats", STATS_WINDOW_SIZE, STATS_NUM_WINDOWS);
 
 FieldMap fm = { { "WTW_0_1", "int" },
-		{ "WTW_1_10", "int" },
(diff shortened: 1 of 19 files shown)


common/perf-stats.cc
```diff
@@ -42,7 +42,9 @@ FieldMap fm = { { "WTW_0_1", "int" },
 		{ "WTR_1_10", "int" },
 		{ "WTR_10_100", "int" },
 		{ "WTR_100_1000", "int" },
-		{ "WTR_Slow", "int" } };
+		{ "WTR_Slow", "int" },
+                { "Command", "countmin"},
```
cmouse: whitespace nit

common/sstuff.hh
```diff
@@ -80,7 +80,12 @@ public:
 
   ~Socket()
   {
-    closesocket(d_socket);
+    try {
+      closesocket(d_socket);
+    }
+    catch (std::runtime_error &e) {
+      // not a lot we can do here
```
cmouse: you could log this.

PR #192, opened 2018-03-15
Geoip2

neilcook:


m4/pdns_with_geo.m4
```diff
@@ -0,0 +1,47 @@
...
+        AC_CHECK_LIB([maxminddb], [MMDB_open], [
+          AC_DEFINE([HAVE_MMDB], [1], [Define this if you have Maxmind DB])
+          MMDB_CFLAGS="-I$with_maxminddb_incdir"
+          MMDB_LIBS="-L$with_maxminddb_libdir -lmaxminddb"
+        ])
+      ])
+    ])
+  ])
+
+  PKG_CHECK_MODULES([YAML], [yaml-cpp >= 0.5],[],
```
cmouse: this is not maxmind!

configure.ac
```diff
@@ -9,7 +9,7 @@ AC_LANG([C++])
 PDNS_CHECK_OS
 # Check dependencies with pkgconfig
 PKG_PROG_PKG_CONFIG
-PKG_CHECK_MODULES([GEOIP], [geoip], [AC_DEFINE([HAVE_GEOIP], [1], [Define to 1 if you have geoip])], [true])
+PDNS_CHECK_GEOIP
```
cmouse: move this commit before the previous one?

m4/pdns_with_geo.m4
neilcook: Indeed, well spotted. I just copied this from pdns, so I guess it has the same problem!

configure.ac
neilcook: yeah good point

m4/pdns_with_geo.m4
cmouse: In PowerDNS the geoip backend uses YAML as config format, which is why this there.

common/wforce-geoip2.cc
```diff
@@ -0,0 +1,143 @@
...
+
+  if (!mmdbLookup(address.toString(), res) ||
+      (MMDB_get_value(&res.entry, &data, "isp", NULL) != MMDB_SUCCESS) ||
+      !data.has_data)
+    return {};
+
+  return std::string(data.utf8_string, data.data_size);
+}
+
+WFGeoIPRecord WFGeoIP2DB::lookupCity(const ComboAddress& address)
```
cmouse: lookupCity? This does not lookup city? =)

common/wforce-geoip2.cc
```diff
@@ -0,0 +1,143 @@
...
+  }
+  else if (res.found_entry) {
+    return true;
+  }
+  return false;
+}
+
+std::shared_ptr<WFGeoIP2DB> WFGeoIP2DB::makeWFGeoIP2DB(const std::string& filename)
+{
+  return std::make_shared<WFGeoIP2DB>(filename);
```
cmouse: why not unique_ptr?

common/wforce-geoip2.cc
neilcook: Well it does lookup city, along with a bunch of other fields, all of which (in combination) are only available in the City DB.

neilcook: Also, I wanted to keep the interface identical to the existing legacy functions (as much as possible)

common/wforce-geoip2.cc
neilcook: Because it has to be shared with Lua

common/wforce-geoip2.cc
neilcook: i.e. the XXX in lookupXXX refers more to the DB that is being looked up in (because that's how the legacy API was written)

cmouse: just looks confusing

neilcook: Well that's the way it is, and I'm really loath to change the API cos there's already customers using it, and I want to make their switch to Geoip2 as easy as possible

Issue #50, opened 2016-04-21
Chaining commands to siblings (for when warming up_


neilcook:

neilcook:
Alternatively when booting, synchronise statsdb old time windows (except current) (MRU first) with a wforce sibling that already has the stats (assuming there is one).

neilcook:
One model:
- Set a "min time sibling must be up" before a newly started server will pull old stats from it
- Optionally set a list of siblings that can only be used for pulling stats
- If I find any siblings which match the above:
    - Don't listen on port
    - Contacts a random sibling (only from the list if it's configured) that has been up long enough, and tell it to send us all old stats (only old time windows not the current time window?)
    - Keep going until we have all the old time windows and all the keys
    - Start server by listening on port, tell systemd we're ready
- Otherwise start up immediately (listen on port, tell systemd we're ready)

neilcook:
Also consider making the standard replication work over TCP as part of this work

neilcook:
Merged #202

Issue #48, opened 2016-04-11
"ping" command for clients to cheaply check weakforce is up


neilcook:

neilcook:
Implement two different results for ping command:
1) Warming up - Only send me report commands
2) Fully Up - Send me report and allow commands

neilcook:
Ping states would only be for wf servers with configured siblings - it makes no sense for a single node

neilcook:
Merged #202

PR #202, opened 2018-05-08
Synchronization of DBs for newly started wforce instances

neilcook:
As per #48 and #50:

- Set a "min time sibling must be up" before a newly started server will pull old stats from it
- Set a list of "sync" servers that will be used for pulling stats
- Contact each configured sync server and see if it has been up long enough
- If wforce finds any catchup servers which have been up long enough:
  - Set ping status to "warming up" - this needs a new ping command response (see #48)
  - Contacts a catchup server that has been up long enough, and tell it to send us all stats (including the current time window)
 - Keep going until we have all the old time windows and all the keys (we know this because we set a callback for the catchup server to tell us when its done)
 - Set ping status to "ready"
- If no catchup servers available which exceed the minimum uptime, then set ping status to "ready"

Also adds replication over TCP using an additional ":tcp" specifier for Sibling addresses.

Note that means the following behaviour changes:
- If no sync servers are defined, or wforce doesn't find any that exceed the minimum uptime, or doesn't find any responding with a "status: ok" response to syncDBs command, then the server will start in the "ready" ping state.
- If a sync server is found, and it responds with "status: ok" then wforce will be in the "warmup" ping state until the callback is received. If the callback URL is misconfigured then wforce will never leave the ping "warmup" state.
- Note that all of this only changes the ping-state; all other commands works as normal including allow and report etc. So this *only* affects deployments where the client (or an intermediary) such as HAProxy uses the ping response to determine whether to send requests to a wforce instance. Currently Dovecot does not use this, although it may in future.
- A server designated as a "sync" server probably shouldn't be a production wforce server. Synchronizing DBs not only places a large load on the server, but it also locks the each Stats DB for the duration of the sync. This could cause delays in responding to allow/report requests.

### DIFF
--- a/common/common-lua.cc
+++ b/common/common-lua.cc
@@ -255,6 +255,13 @@ void setupCommonLua(bool client,
   }
 
   if (!multi_lua) {
+    c_lua.writeFunction("setWebHookTimeoutSecs", [](uint64_t timeout_secs) { g_webhook_runner.setTimeout(timeout_secs); });
+  }
+  else {
+    c_lua.writeFunction("setWebHookTimeoutSecs", [](uint64_t timeout_secs) { });
+  }
+  
+  if (!multi_lua) {
     c_lua.writeFunction("showCustomWebHooks", []() {
 	auto webhooks = g_custom_webhook_db.getWebHooks();
 	boost::format fmt("%-9d %-20.20s %-9d %-9d %-s\n");
--- a/common/sstuff.hh
+++ b/common/sstuff.hh
@@ -236,8 +236,6 @@ public:
       res=::send(d_socket, ptr, toWrite, 0);
       if(res<0) 
         throw NetworkError("Writing to a socket: "+string(strerror(errno)));
-      if(!res)
-        throw NetworkError("EOF on socket");
       toWrite-=res;
       ptr+=res;
     }while(toWrite);
--- a/common/sstuff.hh
+++ b/common/sstuff.hh
@@ -345,6 +345,16 @@ public:
     return res;
   }
 
+  //! Reads a block of data from the socket to a block of memory
+  //! Only returns when all the data requested is available
+  int readAll(char *buffer, int bytes)
+  {
+    int res=::recv(d_socket,buffer,bytes, MSG_WAITALL);
+    if(res<0) 
+      throw NetworkError("Reading from a socket: "+string(strerror(errno)));
+    return res;
+  }
+
   int readWithTimeout(char* buffer, int n, int timeout)
   {
     int err = waitForRWData(d_socket, true, timeout, 0);
--- a/common/twmap.hh
+++ b/common/twmap.hh
@@ -64,6 +64,8 @@ public:
   virtual void erase() = 0; // Remove all data - notice this isn't necessarily the same as set(0) or set("")
   virtual int sum(const TWStatsBuf& vec) = 0; // combine an array of stored values
   virtual int sum(const std::string& s, const TWStatsBuf& vec) = 0; // combine an array of stored values
+  virtual void dump(std::ostream& os) = 0; // serialize a TWStatsMember
+  virtual void restore(std::istream& is) = 0; // unserialize a TWStatsMember
 };
 
 class TWStatsMemberInt : public TWStatsMember
@@ -94,6 +96,18 @@ public:
     return count;
   }
   int sum(const std::string& s, const TWStatsBuf& vec) { return 0; }
+  void dump(std::ostream& os) {
+    uint32_t neti = htonl(i);
+    os.write((char*)&neti, sizeof(neti));
+    if(os.fail()) {
+      throw std::runtime_error("TWStatsMemberInt: Failed to dump");
+    }
+  }
+  void restore(std::istream& is) {
+    uint32_t neti = 0;
+    is.read((char*)&neti, sizeof(neti));
+    i = ntohl(neti);
+  }
 private:
   int i=0;
 };
@@ -132,6 +146,12 @@ public:
     return std::lround(hllsum.estimate());
   }
   int sum(const std::string& s, const TWStatsBuf& vec) { return 0; }
+  void dump(std::ostream& os) {
+    hllp->dump(os);
+  }
+  void restore(std::istream& is) {
+    hllp->restore(is);
+  }
   static void setNumBits(unsigned int nbits) {
     if (nbits < 4)
       nbits = 4;
@@ -186,6 +206,12 @@ public:
       }
     return count;
   }
+  void dump(std::ostream& os) {
+    cm->dump(os);
+  }
+  void restore(std::istream& is) {
+    cm->restore(is);
+  }
   static void setGamma(float g)
   {
     if (g > 1)
@@ -237,6 +263,8 @@ protected:
     type_map["countmin"] = &createInstance<TWStatsMemberCountMin>;
   }
 };
+
+typedef std::vector<std::pair<std::time_t, std::stringstream>> TWStatsBufSerial;
 
 // this class is not protected by mutexes because it sits behind TWStatsDB which has a mutex
 // controlling all access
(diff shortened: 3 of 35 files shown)


neilcook:
TODO: Regression tests and some performance testing of the DB sync

neilcook:
Some basic performance testing showed that around 350K entries took around 10 seconds to synchronise. On a fairly slow Digital Ocean basic VM with 2GB RAM. I couldn't test anymore at that point because I ran out of memory. Each entry wasn't particularly big (only one field), so I expect real-world replication to take somewhat longer.

If we assume that real-world replication takes at least 3x longer (due to the larger number of fields in real-world policies), then the default DB size of 500K entries would take around 45-60 seconds to complete.

The only real downside to replication taking a decent amount of time is that the sending wforce server has to lock each DB in turn while it synchronises. Which means that replication requests and HTTP requests will hang until the sync is finished. This is why I recommend to use "dedicated" sync servers that don't handle allow/report requests, but do participate in the replication. If a few replication events get lost because the UDP buffers fill up while synchronising, this is not a big deal (since those servers are not used for queries).

common/common-lua.cc
```diff
@@ -254,6 +254,13 @@ void setupCommonLua(bool client,
     c_lua.writeFunction("setNumWebHookConnsPerThread", [](unsigned int num_conns) { });
   }
 
+  if (!multi_lua) {
```
cmouse: Somehow I feel it might be a good idea at some point to move all this `!multi_lua` stuff into some separate method/function/whatever that will initialize these instead of the other. Would be cleaner and easier to read.

common/sstuff.hh
```diff
@@ -345,6 +345,16 @@ public:
     return res;
   }
 
+  //! Reads a block of data from the socket to a block of memory
+  //! Only returns when all the data requested is available
+  int readAll(char *buffer, int bytes)
```
cmouse: any particular reason this is using char and int? it could use std::string and size_t instead. also it should probably return ssize_t instead of int.

common/twmap.hh
```diff
@@ -94,6 +96,18 @@ public:
     return count;
   }
   int sum(const std::string& s, const TWStatsBuf& vec) { return 0; }
+  void dump(std::ostream& os) {
```
cmouse: you should add & user 'override' keyword with members that override methods, this way the compiler can tell you if it changes in the future.

common/twmap.hh
```diff
@@ -238,6 +264,8 @@ protected:
   }
 };
 
+typedef std::vector<std::pair<std::time_t, std::stringstream>> TWStatsBufSerial;
+
 // this class is not protected by mutexes because it sits behind TWStatsDB which has a mutex
 // controlling all access
 class TWStatsEntry
```
cmouse: This could be a templated class instead.

common/twmap.hh
```diff
@@ -479,6 +530,50 @@ public:
...
+    mutx.lock();
+    return key_tracker.begin();
+  }
+  // While the mutex is being held, no modifications can be made to the DB;
+  // this could cause replication packets to get backed up and lost
+  bool DBDumpEntry(typename TWKeyTrackerType::iterator& i,
+                   TWStatsDBDumpEntry& entry,
+                   T& key)
+  {
+    const typename TWStatsDBMap::iterator it = stats_db.find(*i);
```
cmouse: const auto?

ext/ext/count_min_sketch.cpp
```diff
@@ -143,4 +145,53 @@ unsigned int CountMinSketch::hashstr(const char *str) {
...
+  std::swap(d, rhs.d);
+  std::swap(C, rhs.C);
+  std::swap(hashes, rhs.hashes);
+}
+
+// XXX - this function does not currently convert everything to network byte order
+void CountMinSketch::dump(std::ostream& os) const throw(std::runtime_error)
+{
+  int i=0;
+  os.write((char*)&eps, sizeof(eps));
```
cmouse: why not use https://www.boost.org/doc/libs/1_59_0/libs/endian/doc/buffers.html instead since you are already using boost?

wforce/wforce-lua.cc
```diff
@@ -176,6 +176,33 @@ vector<std::function<void(void)>> setupLua(bool client, bool multi_lua, LuaConte
...
 
+  if (!multi_lua) {
+    c_lua.writeFunction("addSyncHost", [](const std::string& address, const std::string password, const std::string& sync_address, const std::string& webserver_address) {
+	try {
+          g_sync_data.sync_hosts.push_back(std::make_pair(ComboAddress(address, 8084), password));
+          g_sync_data.sibling_listen_addr = ComboAddress(sync_address, 4001);
+          g_sync_data.webserver_listen_addr = ComboAddress(webserver_address, 8084);
+	}
+	catch (const WforceException& e) {
+          boost::format fmt("%s (%s)\n");
```
cmouse:
```.c++
const std::string errstr = (boost::format("%s (%s)\n") % "addSyncHost(): Error parsing address/port. Make sure to use IP addresses not hostnames", e.reason).str();
errlog(errstr);
g_outputBuffer +=errstr;
```
maybe?

wforce/wforce.cc
```diff
@@ -460,23 +465,329 @@ unsigned int g_num_sibling_threads = WFORCE_NUM_SIBLING_THREADS;
...
+void encryptMsg(const std::string& msg, std::string& packet)
+{
+    std::lock_guard<std::mutex> lock(sod_mutx);
+    packet=g_sodnonce.toString();
+    packet+=sodEncryptSym(msg, g_key, g_sodnonce);
+}
+
+bool decryptMsg(const char* buf, int len, std::string& msg)
+{
+  SodiumNonce nonce;
```
cmouse: maybe make sure len > nnn?

wforce/wforce.cc
```diff
@@ -460,23 +465,329 @@ unsigned int g_num_sibling_threads = WFORCE_NUM_SIBLING_THREADS;
...
+  memcpy((char*)&nonce, buf, crypto_secretbox_NONCEBYTES);
```
cmouse: also if buf was std::string, you could use std::string#copy instead.

common/sstuff.hh
```diff
@@ -345,6 +345,16 @@ public:
...
+  {
+    int res=::recv(d_socket,buffer,bytes, MSG_WAITALL);
```
cmouse: i am not sure if this is a good idea. maybe better idea would be to call ::recv until enough bytes (decrement the bytes with what has been read), until all has been received or error occurs?

common/sstuff.hh
neilcook: Well recv needs char* buffer. If I pass a std::string into the function then I'd have to dynamically allocate a char* buffer of the right size in the function to use with recv, and then copy the result into the string. That doesn't seem very efficient.

common/twmap.hh
neilcook: yep good point

common/twmap.hh
neilcook: yep

wforce/wforce-lua.cc
neilcook: maybe what? :)

common/sstuff.hh
neilcook: I could, but how would that be different from using MSG_WAITALL, which essentially implements exactly the same algorithm.

wforce/wforce-lua.cc
neilcook: oh I see, yes

wforce/wforce.cc
neilcook: buffer is coming directly off a socket read so I don't want to have to convert it to std::string

common/sstuff.hh
neilcook: Also the int stuff is simply to match the other functions in this class. I should probably change them all if I'm going to change one

wforce/wforce.cc
neilcook: yep

common/common-lua.cc
cmouse: you have quite a lot of those `if (!multi_lua) {` branches here. you could have `setupCommonBase´, `setupCommonMultiLua` and `setupCommonLua` instead.

common/twmap.hh
cmouse: It would reduce code duplication I guess.

common/sstuff.hh
cmouse: hmm. MSG_WAITALL can return less than bytes. It does not guarantee that res = bytes.

neilcook:
I also just added TCP replication to this PR, since the rest of PR did most of the work, and Softbank have asked for both features

neilcook:
Let me know if you have any more comments/suggestions, otherwise I'll merge this

PR #204, opened 2018-05-18
Improve web server connection handling

neilcook:
This PR improves the connection handling of the webserver under high load conditions, or where there are a high number of connections. Previously when the max number of active connections was reached (5000 was hardcoded), connections would be not be added to the ctpl thread pool queue, they would be silently dropped. They would then enter a non-responsive zombie state until the client closed the connection.
Now an explicit queue is used, and if the queue gets too big (max conns) then even if there is activity on the socket from poll, that connection won't be added to the queue until the queue has room. Also the max number of active connections is now configurable from Lua.

This PR was inspired by issues at Comcast, where they fairly regularly have over 5000 connections to each wforce instance, and saw a lot of timeout issues.


common/wforce-webserver.cc
```diff
@@ -379,9 +381,10 @@ void WforceWebserver::pollThread(WforceWebserver* wws)
               wws->d_queue.push(wqi);
             }
           }
-          wws->d_cv.notify_all();
 	}
       }
+      if (pollin)
+        wws->d_cv.notify_all();
```
cmouse: why not just `if (wws->d_queue.size() > 0)`

neilcook: Hmm that's a good point. Will make that change I think

PR #208, opened 2018-06-25
Fixes

neilcook:


common/minicurl.cc
```diff
@@ -198,8 +198,8 @@ bool MiniCurl::postURL(const std::string& url,
     else {
       long response_code;
       curl_easy_getinfo(d_curl, CURLINFO_RESPONSE_CODE, &response_code);
-      if (response_code != 200) {
-	error_msg = std::string("Received non-200 response from webserver: ") + std::to_string(response_code);
+      if ((response_code < 200) || (response_code > 299)) {
```
cmouse: why not just do `if (response_code / 100 != 2)`

neilcook: Well yes, that also possible, but I don't think there's a particular reason to change it is there?

cmouse: Perhaps. I just wonder if miniCurl should have something like `bool is2xx(const int& code) { return code/100 == 2; }`.cc instead?

neilcook: Well that is a very good idea

PR #233, opened 2018-10-26
Add sanitizers support and enable them for CI tests

chbruyand:


chbruyand:
g++ version is quiet old. And it looks like the clang regression test is not really using clang :
```
$ export CXX=g++
$ export CC=gcc
```

neilcook:
> g++ version is quiet old. And it looks like the clang regression test is not really using clang :
> 
> ```
> $ export CXX=g++
> $ export CC=gcc
> ```

So regression isn't running in travis directly, it runs in docker. So the travis-defined CC/CXX is irrelevant; instead the build matrix sets COMPILER=gcc or COMPILER=clang and then that is passed to the regression tests docker image.

neilcook:
As travis failing, I'm assuming this is WIP for now.

chbruyand:
Looks like it works Travis is happy. Please note that it takes 5 mins longer to run compared to other PRs.

PR #239, opened 2018-11-06
Fix compile to work

cmouse:
- GeoIP was not correctly treated as optional component
 - lua.hpp check was broken


neilcook:
Looks good, thanks

PR #260, opened 2019-04-04
epel: install /usr/bin/python3 for build-requires-from-specs

Habbie:
Based on https://github.com/PowerDNS/pdns/pull/7671, but bigger, because you actually use Python for more than the build process. It looks like the newer `pip`(?) is stricter about requirements.txt syntax, so I updated that too.

I only checked that `build.sh centos-7` passes with this change (which, to be clear, it no longer does without this change, see https://www.mail-archive.com/epel-devel@lists.fedoraproject.org/msg06502.html). I did not verify any functionality.

# Open questions/problems

`builder-support/dockerfiles/Dockerfile.rpmbuild:RUN yum install -y rpm-build rpmdevtools python3-rpm-macros python34 python34-pip python34-devel && \` - if you want to make these version agnostic, I think you want `/usr/bin/python3 /usr/bin/pip3 /usr/bin/python3-config` instead of the last 3 package names. This will, today, give you the `36` variants of the three, and should be stable for any future Python 3 upgrades in EPEL.

`builder-support/specs/wforce.spec:Requires: python34` - I did not check whether the resulting package actually works.


Habbie:
A friendly person on #epel notes that you can use `%{python3_pkgversion}` in your spec-file.

Habbie:
Travis failed - I've added a commit to show `test-suite.log` to get more insight.

Habbie:
It passed this time. I'll leave the commit in - it is helpful for next time.

neilcook:
Thanks Peter. I'll also test with debian. I really should add the packaging stuff to the travis matrix.

Issue #261, opened 2019-04-10
Query Blacklist for firewall integration


sshipway:
**Is your feature request related to a problem? Please describe.**
The pfSense firewall (and other firewalls) has a feature where it can blacklist incoming connections based on a list of IP addresses retrieved by a URL.  It would be very helpful for us for incoming blacklisted connections to be able to be dropped at the firewall rather than having to attempt login.

**Describe the solution you'd like**
If Weakforced were able to support a query URL that returns a plain list of IP addresses, one per line, that are valid in the blacklist, then this could be hooked directly into by pfSense.

For example, if we could call a URL such as
  http://weakforced/blacklist?minvalid=300&format=plain
to extract a plain-text list of IP addresses, one per line, that have at least 300s validity remaining in the blacklist (this would be so we could synch every 300s to minimise load).

**Describe alternatives you've considered**
Since the blacklist data are already available on the /command=getBL URL, we have tried using this.  However, this comes in JSON format and needs to be parsed, filtered, reformatted and checked for time remaining, and would require us to create a further CGI script on a webserver to convert, or else a whole plugin for the firewall.  Both solutions seem to be suboptimal.

**Additional context**
The more integrations that are natively supported by weakforced, the more useful it becomes.  Using firewalls to protect against brute-force bad actor IP addresses is desirable, but most firewalls have a very simple set of formats supported (DNSBL or HTTP plain text seem to be the usual).
This should be a relatively simple extension to the existing getBL functionality.
A further extension that would aid compatibility would be to be able to optionally disable authentication for this particular query, as some firewalls do not support this.

neilcook:
So there is the custom endpoint command:

setCustomEndpoint(<name  of  endpoint>,  <boolean>, <custom lua func-
         tion>) - Create a new custom REST endpoint with the given name, which
         when invoked will call the supplied custom lua function.  This allows
         admins to arbitrarily extend the wforce REST API with new  REST  end-
         points.   Admins can create as many custom endpoints as they require.
         Custom endpoints can only be accessed via a POST method, and all  ar-
         guments  as passed as key-value pairs of a top-level "attrs" json ob-
         ject (these will be split into two tables - one for single-valued at-
         trs,  and  the  other for multi-valued attrs - see CustomFuncArgs be-
         low).  If the boolean argument is true, then  all  arguments  to  the
         custom  function  will  be sent to all configured named report sinks.
         Return information is passed with a boolean "success"  and  "r_attrs"
[...]

PR #268, opened 2019-05-21
Custom GET Endpoint

neilcook:
- Custom GET Endpoint (see #261)
- New Lua functions to retrieve the various black/whitelists in their entirety
- Use %{python3_version} in the report_api config rather than hardcode it


sshipway:
The Get endpoint combined with the function to retrieve black/whitelists will solved #261

common/wforce-webserver.cc
```diff
@@ -71,7 +71,7 @@ size_t WforceWebserver::getNumConns()
   return d_sock_vec.size();
 }
 
-bool WforceWebserver::registerFunc(const std::string& command, HTTPVerb verb, WforceWSFunc func)
+bool WforceWebserver::registerFunc(const std::string& command, HTTPVerb verb, const WforceWSFunc& func)
```
cmouse: unrelated change.

common/wforce-webserver.cc
```diff
@@ -224,17 +224,21 @@ void WforceWebserver::connectionThread(WforceWebserver* wws)
         // set the defaults in case we don't find a command
         resp.status = 404;
         resp.body = R"({"status":"failure", "reason":"Command not found"})";
-    
+
+        std::string new_content_type;
```
cmouse: totally unrelated change.

neilcook: Why is this unrelated? The purpose of this new function is to be able to return different content types from the GET functions. Not also allowing that for the other types of functions would be reverse

common/wforce-webserver.cc
neilcook: will break out into a single commit

common/wforce-webserver.cc
cmouse: i cannot see how it's related to the commit, so perhaps you should do this as a preparation before this commit?

neilcook: ok

PR #271, opened 2019-05-27
Geoip2 lookup individual values

neilcook:
This PR implements Lua functions to lookup individual values, to allow any GeoIP DB to be interrogated.

### DIFF
--- a/common/wforce-geoip2.cc
+++ b/common/wforce-geoip2.cc
@@ -25,6 +25,7 @@
 #include "wforce-geoip2.hh"
 #include "wforce_exception.hh"
 #include "dolog.hh"
+#include <string.h>
 
 std::mutex geoip2_mutx;
 std::map<std::string, std::shared_ptr<WFGeoIP2DB>> geoip2Map;
@@ -137,9 +138,7 @@ bool WFGeoIP2DB::mmdbLookup(const std::string& ip, MMDB_lookup_result_s& res)
 
 char* convertToCharStar(const std::pair<unsigned int, std::string>& pair)
 {
-  char *cs = new char[pair.second.size()+1];
-  std::strcpy(cs, pair.second.c_str());
-  return cs;
+  return strdup(pair.second.c_str());
 }
 
 bool WFGeoIP2DB::lookupDataValue(const ComboAddress& address, const std::vector<std::pair<unsigned int, std::string>>& attrs, MMDB_entry_data_s& ret_data)
@@ -156,7 +155,7 @@ bool WFGeoIP2DB::lookupDataValue(const ComboAddress& address, const std::vector<
     }
     for (auto& i : path) {
       if (i != nullptr)
-        delete i;
+        free(i);
     }
   }
   return retval;
--- a/common/wforce-geoip2.cc
+++ b/common/wforce-geoip2.cc
@@ -154,8 +154,7 @@ bool WFGeoIP2DB::lookupDataValue(const ComboAddress& address, const std::vector<
       retval = true;
     }
     for (auto& i : path) {
-      if (i != nullptr)
-        free(i);
+      free(i);
     }
   }
   return retval;
--- a/common/wforce-geoip2.cc
+++ b/common/wforce-geoip2.cc
@@ -135,6 +135,93 @@ bool WFGeoIP2DB::mmdbLookup(const std::string& ip, MMDB_lookup_result_s& res)
   return false;
 }
 
+char* convertToCharStar(const std::pair<unsigned int, std::string>& pair)
+{
+  char *cs = new char[pair.second.size()+1];
+  std::strcpy(cs, pair.second.c_str());
+  return cs;
+}
+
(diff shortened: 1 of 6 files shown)


common/wforce-geoip2.cc
```diff
@@ -156,7 +156,7 @@ bool WFGeoIP2DB::lookupDataValue(const ComboAddress& address, const std::vector<
     }
     for (auto& i : path) {
       if (i != nullptr)
-        delete i;
+        delete[] i;
```
cmouse: iirc delete[] is NULL safe.

common/wforce-geoip2.cc
```diff
@@ -135,6 +135,93 @@ bool WFGeoIP2DB::mmdbLookup(const std::string& ip, MMDB_lookup_result_s& res)
   return false;
 }
 
+char* convertToCharStar(const std::pair<unsigned int, std::string>& pair)
+{
+  char *cs = new char[pair.second.size()+1];
+  std::strcpy(cs, pair.second.c_str());
+  return cs;
```
cmouse: `return std::strdup(pair.second.c_str())`?

common/wforce-geoip2.cc
```diff
@@ -135,6 +135,93 @@ bool WFGeoIP2DB::mmdbLookup(const std::string& ip, MMDB_lookup_result_s& res)
...
+  
+  if (mmdbLookup(address.toString(), res)) {
+    std::transform(attrs.begin(), attrs.end(), std::back_inserter(path), convertToCharStar);
+    path.emplace_back(nullptr);
+    if ((MMDB_aget_value(&res.entry, &ret_data, &path[0]) == MMDB_SUCCESS) && (ret_data.has_data)) {
+      retval = true;
+    }
+    for (auto& i : path) {
+      if (i != nullptr)
+        delete i;
```
cmouse: delete is null safe, also you probably should use delete[] here?

common/wforce-geoip2.cc
neilcook: Yeah that would also work. Also I guess I'd need to use free() rather than delete[] (and keep the check for nullptr)

common/wforce-geoip2.cc
neilcook: Weirdly I only added the check because I got a SEGV otherwise

common/wforce-geoip2.cc
```diff
@@ -156,7 +155,7 @@ bool WFGeoIP2DB::lookupDataValue(const ComboAddress& address, const std::vector<
     }
     for (auto& i : path) {
       if (i != nullptr)
-        delete[] i;
+        free(i);
```
cmouse: free is null-safe too.

neilcook: well I never!

common/wforce-geoip2.cc
cmouse: no free here?

docker/Makefile.am
```diff
@@ -12,7 +12,7 @@ $(GEOIP_FILE_GZ):
 $(GEOIP_FILE): $(GEOIP_FILE_GZ)
 	gunzip -c $(GEOIP_FILE_GZ) >$(GEOIP_FILE)
 
-$(COMPOSE_TARGET): $(COMPOSE_SOURCE) $(GEOIP_FILE) $(shell find ../common -type f) $(shell find ../wforce -type f) $(shell find ../trackalert -type f) $(shell find ../ext -type f) 
+$(COMPOSE_TARGET): $(COMPOSE_SOURCE) $(GEOIP_FILE) $(shell find ../common -type f) $(shell find ../wforce -type f) $(shell find ../trackalert -type f) $(shell find ../ext -type f) $(shell find ../regression-tests -type f)
```
cmouse: why is this here?

common/wforce-geoip2.cc
```diff
@@ -156,7 +155,7 @@ bool WFGeoIP2DB::lookupDataValue(const ComboAddress& address, const std::vector<
     }
     for (auto& i : path) {
       if (i != nullptr)
-        delete i;
+        free(i);
```
cmouse: this could already be done in previous commit instead of first doing it there and then changing it here.

common/wforce-geoip2.cc
```diff
@@ -137,9 +138,7 @@ bool WFGeoIP2DB::mmdbLookup(const std::string& ip, MMDB_lookup_result_s& res)
 
 char* convertToCharStar(const std::pair<unsigned int, std::string>& pair)
 {
-  char *cs = new char[pair.second.size()+1];
-  std::strcpy(cs, pair.second.c_str());
-  return cs;
+  return strdup(pair.second.c_str());
```
cmouse: this could already be done in the earlier commit.

common/wforce-geoip2.cc
```diff
@@ -154,8 +154,7 @@ bool WFGeoIP2DB::lookupDataValue(const ComboAddress& address, const std::vector<
       retval = true;
     }
     for (auto& i : path) {
-      if (i != nullptr)
-        free(i);
+      free(i);
```
cmouse: this could already happen in earlier commit.

PR #275, opened 2019-06-10
Blacklist Improvements

neilcook:
Getting the blacklist return message from Lua is essential in order to implement Lua-based blacklist checks in policy.


neilcook:
thx!

PR #278, opened 2019-08-12
Code Cleanup/Refactor

neilcook:
Reviewing the code for a possible memory leak at Comcast found a few issues including:
- StatsDB expiry was broken when running with -d (demonised) - thread initialisation needed to be moved after the fork() rather than in the constructor
- The way the replication thread pool was initialised was not actually thread-safe, so this has been moved to the main initialisation section, which is also cleaner and removes the need for it to be thread-safe.

There's also a cleanup of the replication destructors which isn't functional but rather to make it cleaner:
- Use default rather than {} in the virtual destructor
- Remove the child class destructors (not all of them had destructors anyway) as they were just {}

Finally fix a deadlock bug introduced by not scoping the dbMap mutex correctly.

### DIFF
--- a/wforce/replication.hh
+++ b/wforce/replication.hh
@@ -31,7 +31,7 @@ typedef std::shared_ptr<AnyReplicationOperation> AnyReplicationOperationP;
 class AnyReplicationOperation
 {
 public:
-  virtual ~AnyReplicationOperation() {}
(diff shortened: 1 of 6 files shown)


neilcook:
Ok I modified the commit messages

PR #287, opened 2019-10-28
Prometheus support for weakforced

neilcook:
This PR adds prometheus support to weakforced on the /metrics endpoint of both the wforce and trackalert daemons.

This PR includes:
- Documentation updates (Man pages and OpenAPI specs)
- Regression tests

All the prometheus support is squashed into a single commit.

There are several other commits which fix small issues found during the prometheus support development, notably:
- Make libcrypto a non-optional dependency
- More intuitive constructors for WFResolver class
- Handle zero-length encrypted data more safely
- Fix syncDBs stats name for deprecated stats method
- Fix wrong prototype for empty runCustomWebHook() Lua function


common/prometheus.hh
```diff
@@ -0,0 +1,185 @@
...
+    webhook_queue_size->Set(value);
+  }
+
+  std::string serialize()
+  {
+    return d_textSerializer.Serialize(d_registry->Collect());
+  }
+
+private:
+  Histogram* worker_queue_duration;
```
chbruyand: At a quick glance, it looks strange to use raw pointers everywhere.

common/prometheus.cc
```diff
@@ -0,0 +1,216 @@
...
+{
+  Counter* c = &(custom_family->Add({{"metric", name}}));
+  custom_metrics.insert(std::make_pair(name, c));
+}
+
+void PrometheusMetrics::incCustomMetric(const std::string& name)
+{
+  auto i = custom_metrics.find(name);
+  if (i != custom_metrics.end()) {
+    i->second->Increment();
```
chbruyand: Not sure this is thread safe

neilcook: You mean the Increment? That class is thread-safe: https://jupp0r.github.io/prometheus-cpp/classprometheus_1_1Counter.html

common/prometheus.hh
neilcook: That's because these objects are created and managed by the prometheus Registry class. It returns a reference to the object from BuildXXXX().

neilcook: If you look at Registry.Add() you'll see that the object is managed with a unique ptr

neilcook:
This is failing due to the max mind changes, which are fixed in #294, so they should be merged first

PR #289, opened 2019-11-14
Replication fixes

neilcook:
This PR started as a cleanup of the sibling and replication code into separate source files for better code reuse and future projects involving replication.

Testing this altered me to a lurking bug in the syncDB code where a thread created by a Sibling instance would cause a crash if the instance is destructed (this normally doesn't happen because Siblings tend to hang around for the lifetime of the process). It turns out that the Sibling allocated on the stack for the syncDB code was not being used, so fixing the crash bug was a matter of removing it completely. However the Sibling code still needed fixing, so now the Sibling destructor explicitly tells the child thread to stop, and joins it to prevent std::terminate from being called when the thread destructor is called.

While testing the crash I noticed that the wforce client was also exhibiting this behaviour, which led me to discover that Siblings and Report Sinks are being created in the client. This no longer happens.

This PR will likely form the basis for 2.2.1 because syncDB is not reliably usable without this fix.


wforce/wforce-replication.hh
```diff
@@ -0,0 +1,44 @@
+/*
+ * This file is part of PowerDNS or weakforced.
+ * Copyright -- PowerDNS.COM B.V. and its contributors
```
cmouse: PowerDNS.COM B.V. is still legit here?

wforce/wforce-lua.cc
```diff
@@ -281,7 +282,7 @@ vector<std::function<void(void)>> setupLua(bool client, bool multi_lua, LuaConte
     c_lua.writeFunction("setSiblings", [](const vector<pair<int, string>>& parts) { });
   }
 
-  if (!multi_lua) {
+  if (!multi_lua && !client) {
```
cmouse: maybe || instead?

wforce/wforce-replication.cc
```diff
@@ -0,0 +1,281 @@
...
+struct SiblingQueueItem {
+  std::string msg;
+  ComboAddress remote;
+  std::shared_ptr<Sibling> recv_sibling;
+};
+
+static std::mutex g_sibling_queue_mutex;
+static std::queue<SiblingQueueItem> g_sibling_queue;
+static std::condition_variable g_sibling_queue_cv;
+size_t g_max_sibling_queue_size = 1000;
```
cmouse: const

wforce/wforce-replication.cc
```diff
@@ -0,0 +1,281 @@
...
+    std::lock_guard<std::mutex> lock(sod_mutx);
+    packet=g_sodnonce.toString();
+    packet+=sodEncryptSym(msg, g_key, g_sodnonce);
+}
+
+bool decryptMsg(const char* buf, size_t len, std::string& msg)
+{
+  SodiumNonce nonce;
+
+  if (len < static_cast<int>(crypto_secretbox_NONCEBYTES)) {
```
cmouse: len is size_t, not int.

wforce/wforce-replication.cc
```diff
@@ -0,0 +1,281 @@
...
+  g_sibling_queue_cv.notify_one();
+}
+
+void parseTCPReplication(std::shared_ptr<Socket> sockp, const ComboAddress& remote, std::shared_ptr<Sibling> recv_sibling)
+{
+  setThreadName("wf/repl-tcp");
+
+  infolog("New TCP Replication connection from %s", remote.toString());
+  uint16_t size;
+  unsigned char ssize = sizeof(size);
```
cmouse: size_t?

wforce/wforce-replication.cc
```diff
@@ -0,0 +1,281 @@
...
+      thread t1(parseTCPReplication, connp, remote, recv_sibling);
+      t1.detach();
+    }
+    catch(std::exception& e) {
+      errlog("receiveReplicationOperationsTCP: error accepting new connection: %s", e.what());
+    }
+  }
+}
+
+void receiveReplicationOperations(ComboAddress local)
```
cmouse: &local

cmouse: and could be also const ComboAddress &local

wforce/wforce-sibling.hh
```diff
@@ -0,0 +1,77 @@
...
+  void queueMsg(const std::string& msg);
+  void checkIgnoreSelf(const ComboAddress& ca);
+  void connectSibling();
+  static Protocol stringToProtocol(const std::string& s) {
+    if (s.compare("tcp") == 0)
+      return Sibling::Protocol::TCP;
+    else
+      return Sibling::Protocol::UDP;
+  }
+  static std::string protocolToString(Protocol p) {
```
cmouse: const Protocol &p?

wforce/wforce-replication.hh
neilcook: Yeah still legit. That's who I work for.

wforce/wforce-lua.cc
neilcook: Yeah maybe. I personally find !(xx || y) less readable than (!x && !y) but I think that's just how my brain works (i.e. strangely).

wforce/wforce-replication.cc
neilcook: yep

wforce/wforce-lua.cc
cmouse: well, no, i mean it seems like this should be

if not multi_lua or not client, then....

now it is

if not multi_lua *and* not client then...

neilcook: Ah yes, very good point! :)

neilcook: Also, Doh!

neilcook: So I've thought about this some more and we only want to define the non-stub function for the case where wforce is not a client AND multi-lua is not defined. So the original logic is correct.

neilcook:
Should be good now

PR #294, opened 2020-01-17
Performance Improvements

neilcook:
Three main improvements, all coming from experience at a very large customer with 30 million email subscribers:
1) Make the default statsdb expire/expunge thread sleep much shorter (30 secs -> 250ms), and also configurable. The 30s expiry sleep was causing noticeable spikes in latency causing sibling replication worker thread queues to fill up (this was the visible sign of the underlying issue). Doing expire more frequently means that most requests will take very slightly longer, but will avoid the big latency spikes every 30 seconds.
2) Make the sibling replication worker thread queue size configurable
3) Add an option to make stats dbs sharded. This creates multiple underlying stats dbs underneath the "hood" fo a single Lua statsdb. By doing so, it removes the single-lock contention and should make it much more scalable.

This PR includes regression tests and documentation for new config options

### DIFF
--- a/wforce/twmap-wrapper.cc
+++ b/wforce/twmap-wrapper.cc
@@ -313,6 +313,11 @@ void TWStringStatsDBWrapper::set_size_soft(unsigned int size)
   sdbp->set_map_size_soft(size);
 }
 
+void TWStringStatsDBWrapper::set_expire_sleep(unsigned int ms)
+{
+  sdbp->set_expire_sleep(ms);
+}
+
 unsigned int TWStringStatsDBWrapper::get_max_size()
 {	
   return sdbp->get_max_size();
--- a/wforce/twmap-wrapper.cc
+++ b/wforce/twmap-wrapper.cc
@@ -25,17 +25,26 @@
 #include "replication_sdb.hh"
 #include "replication.pb.h"
 
-TWStringStatsDBWrapper::TWStringStatsDBWrapper(const std::string& name, int window_size, int num_windows)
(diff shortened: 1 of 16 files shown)


wforce/twmap-wrapper.cc
```diff
@@ -100,11 +120,11 @@ std::string TWStringStatsDBWrapper::getStringKey(const TWKeyType vkey)
   else if (vkey.which() == 2) {
     const ComboAddress ca =  boost::get<ComboAddress>(vkey);
     if (ca.isIpv4()) {
-      uint8_t v4_prefix = sdbp->getv4Prefix();
+      uint8_t v4_prefix = sdbvp->at(0)->getv4Prefix();
```
chbruyand: Don't you want to use the computed shard index ?

wforce/wforce.cc
```diff
@@ -505,39 +505,41 @@ void syncDBThread(const ComboAddress& ca, const std::string& callback_url,
...
+              string msg = rep_op.serialize();
+              string packet;
+              encryptMsg(msg, packet);
+              uint16_t nsize = htons(packet.length());
+              rep_sock.writen(std::string((char*)&nsize, sizeof(nsize)));
+              rep_sock.writen(packet);
+              num_synced++;
+            }
+          }
+          catch(const NetworkError& e) {
```
chbruyand: `NetworkError` extends `std::exception`, so maybe you could simplify this ?

wforce/twmap-wrapper.cc
neilcook: Well all the shards will have the same prefix so it doesn't really matter

wforce/wforce.cc
neilcook: yeah, in fact because I do the same thing in all cases, this could be a single catch

PR #298, opened 2020-03-05
Dump Entries Functionality

neilcook:
New functionality to allow the stats DBs to be dumped to a file for debugging purposes (or other reasons).

This consists of a new REST API endpoint that sends all the dumped entries to a specified IP address and port, and a new command line tool "wf_dump_entries" that uses the new API.

The reason this doesn't just return the entries in the REST API response is twofold:
1) To do so would require allocating a giant data structure to hold everything which is unfeasible
2) It would lock up a REST API worker thread for a very long time.

This PR also updates the CHANGELOG with all changes since 2.2.1

### DIFF
--- a/wforce/dump_entries.cc
+++ b/wforce/dump_entries.cc
@@ -136,6 +136,7 @@ int main(int argc, char** argv)
   SSetsockopt(listen_sock, SOL_SOCKET, SO_REUSEADDR, 1);
   SBind(listen_sock, local_ca);
   SListen(listen_sock, 1024);
+  Socket listen_sock_wrapper(listen_sock); // This ensures the socket gets closed
 
   // Connect to the specified wforce address and port
   // and request to dump entries to our local address and port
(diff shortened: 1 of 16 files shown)


wforce/dump_entries.cc
```diff
@@ -0,0 +1,190 @@
...
+  ComboAddress local_ca;
+  try {
+    local_ca = ComboAddress(local_addr, local_port);
+  }
+  catch (const WforceException& e) {
+    cerr << "Error parsing local address and port " << local_addr << ", " << local_port << ". Make sure to use IP addresses not hostnames" << endl;
+    exit(EXIT_FAILURE);
+  }
+  
+  int listen_sock = socket(local_ca.sin4.sin_family, SOCK_STREAM, 0);
```
chbruyand: Is it closed ?

neilcook: Hmm, not by the looks of it.

neilcook:
Yeah I might want to limit concurrent dumps to 1 by default, and make it configurable

wforce/dump_entries.cc
neilcook: Wrapped in a Socket class to auto-close it

neilcook:
> Yeah I might want to limit concurrent dumps to 1 by default, and make it configurable

I made it only one dump at a time - I don't think more than one is a good idea.

neilcook:
This should be good to go now.

PR #299, opened 2020-03-06
Support Kafka REST Proxy in webhooks

neilcook:
Certain folks like to use Kafka as a front-end for all events. Rather than support native kafka, which is a lot of work, this PR supports sending to the Kafka REST Proxy via webhooks, which only requires slightly modifying the webhook content-type and wrapping the json in the format required by Kafka.

This also adds Kafka to the regression tests using the supported Kafka Docker images.


neilcook:
Thanks Max!

Issue #304, opened 2020-04-13
[FEATURE] setCustomGetEndpoint should set the Last-Modified header to the current time


sshipway:
**Is your feature request related to a problem? Please describe.**
In order to use wforced as a IP blacklist source for pfBlockerNG, it needs to provide a list of IP addresses, which it can do.  However it needs to have a Last-Modified header which is not provided, and so the data are rejected.

**Describe the solution you'd like**
The HTTP returned should have the Last-Modified header set to the current time.

**Describe alternatives you've considered**
As this is specifically for pfBlocker integration, the only workaround has been to have a separate cronjob retrieving the data and placing it into a local file to be picked up.

**Additional context**
It might make sense for all returned HTTP responses from any call to include this header

sshipway:
In fact, it might be best to set Date and Last-Modified both to the same value (IE 'now') for optimum compatibility.

neilcook:
Yeah, this seems pretty doable

neilcook:
We should add this header for all REST API responses I think

sshipway:
I agree; although the Date and Last-Modified are defined as optional, there may be other integrations that expect them to be present, and it shouldn't be too much trouble to add them to everything (hopefully)

It might be possible to allow the customGetEndpoint function to optionally return a hash of arbitrary headers along with the content?

PR #308, opened 2020-04-17
Add Last-Modified and Date headers to all HTTP responses

neilcook:
See #304

### DIFF
--- a/common/wforce-webserver.cc
+++ b/common/wforce-webserver.cc
@@ -212,11 +212,13 @@ void WforceWebserver::connectionThread(WforceWebserver* wws)
       char datebuf[WFORCE_MAX_DATE_STRING_LEN];
       time_t now = time(0);
       struct tm tm = *gmtime(&now);
-      std::strftime(datebuf, sizeof datebuf, "%a, %d %b %Y %H:%M:%S GMT", &tm);
+      auto datelen = std::strftime(datebuf, sizeof datebuf, "%a, %d %b %Y %H:%M:%S GMT", &tm);
 
       YaHTTP::Response resp;
-      resp.headers["Date"] = datebuf;
-      resp.headers["Last-Modified"] = datebuf;
+      if (datelen != 0) {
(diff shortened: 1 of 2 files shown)


common/wforce-webserver.cc
```diff
@@ -203,7 +204,15 @@ void WforceWebserver::connectionThread(WforceWebserver* wws)
 
       req.getvars.erase("_"); // jQuery cache buster
 
+      // Generate HTTP Date format string
+      char datebuf[WFORCE_MAX_DATE_STRING_LEN];
```
chbruyand, line 208:
```suggestion
      char datebuf[WFORCE_MAX_DATE_STRING_LEN + 1];
```

common/wforce-webserver.cc
```diff
@@ -203,7 +204,15 @@ void WforceWebserver::connectionThread(WforceWebserver* wws)
...
+      time_t now = time(0);
+      struct tm tm = *gmtime(&now);
+      std::strftime(datebuf, sizeof datebuf, "%a, %d %b %Y %H:%M:%S GMT", &tm);
```
chbruyand, line 211:
```suggestion
      std::strftime(datebuf, sizeof(datebuf) - 1, "%a, %d %b %Y %H:%M:%S GMT", &tm);
```

neilcook: Well the strftime man page says that the \0 is taken into account:
````
 No more than maxsize characters will be placed into the array.  If the total number of resulting charac-
     ters, including the terminating NUL character, is not more than maxsize, strftime() returns the number
     of characters in the array, not counting the terminating NUL.  Otherwise, zero is returned and the
     buffer contents are indeterminate.
````

neilcook: However that does also make me think that I should at least zero out the buffer if strftime returns 0.

chbruyand: > Well the strftime man page says that the \0 is taken into account:

Indeed. Sorry

PR #314, opened 2020-05-18
Redis Keepalive and Configurable Timeout

### LINKED ISSUE #313: [FEATURE] Add keepalive packets to redisdb connection
**Is your feature request related to a problem? Please describe.**
When running redis behind a load balancer (for resiliance) idle connections can occasionally be dropped, and wfd need to reconnect.  In particular, when running in Azure the Azure load balancer silently drops connections without a reset resulting in Weakforced threads hanging as they time out trying to talk to redis.  This then results in all threads being consumed and the service hanging.

**Describe the solution you'd like**
There should be an optional keep-alive packet sent to redis from wforce on a configurable interval, so that any connections via a load balancer can be prevented from idle timeouts

**Describe alternatives you've considered**
There is an option in Azure load balancer to cause a polite shutdown of an idle connection rather than dropping packets.  This is not the default though and can only be set via the API, not via the web interface.  It would be better to have connections remain up and not need to reconnect.

**Additional context**
A separate issue is that the timeout on a redis request seems to be unreasonably long, so that when this problem hits, threads hang for long enough to use up all available threads.

neilcook:
This PR adds the following to the redis black/whitelist capability:

1) Enable TCP keepalive for redis connections
2) Add a configurable timeout for reads/writes on redis connections. Defaults to 0 seconds and 100,000 microseconds.

Fixes #313

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - New wf_dump_entries tool to dump stats DBs to file
 - Support for new "forwarding" type in replication messages
 - Support for Prometheus via the new /metrics REST endpoint
+- Enable TCP keepalive for redis connections
+- Configurable timeout for R/W on redis connections
 
 ### Changed
 - Fix duplicate command stats under some circumstances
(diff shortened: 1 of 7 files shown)


neilcook:
thx!

PR #323, opened 2021-02-09
Add support for building centos-8 and Amazon Linux 2 packages

neilcook:


builder-support/dockerfiles/Dockerfile.target.centos-8
```diff
@@ -0,0 +1,17 @@
...
+RUN yum install -y epel-release
+RUN dnf -y install dnf-plugins-core
+RUN dnf config-manager --set-enabled powertools
+
+# Do the actual rpm build
+@INCLUDE Dockerfile.rpmbuild
+
+# Do a test install and verify
+# Can be skipped with skiptests=1 in the environment
+#@EXEC [ "$skiptests" = "" ] && include Dockerfile.rpmtest
```
Habbie, line 16: Is this commented on purpose?

.travis.yml
```diff
@@ -78,7 +92,6 @@ jobs:
       - cd ..
       - bash docker_push.sh
       - cd ..
-      - ./trigger_policy_build.sh
```
Habbie, line 81: Is this removed on purpose?

builder-support/dockerfiles/Dockerfile.target.amazon-2
```diff
@@ -0,0 +1,14 @@
+# First do the source builds
+@INCLUDE Dockerfile.target.sdist
+
+# This defines the dstribution base layer
```
Habbie, line 4:
```suggestion
# This defines the distribution base layer
```

builder-support/dockerfiles/Dockerfile.target.centos-8
```diff
@@ -0,0 +1,17 @@
+# First do the source builds
+@INCLUDE Dockerfile.target.sdist
+
+# This defines the dstribution base layer
```
Habbie, line 4:
```suggestion
# This defines the distribution base layer
```

builder-support/dockerfiles/Dockerfile.wforce
```diff
@@ -1,5 +1,5 @@
-FROM ubuntu:bionic as wforce
-
+FROM ubuntu:focal as wforce
```
Habbie, line 1: Any reason you're using Ubuntu? For pdns/rec/dnsdist we specifically picked Debian.

builder-support/dockerfiles/Dockerfile.target.centos-8
neilcook: I just based it on the Centos-7 target Dockerfile, so the comments are from there. So no I didn't add the comments.

.travis.yml
neilcook: Yes and no. It should be in its own commit, which I will fix

builder-support/dockerfiles/Dockerfile.wforce
neilcook: I don't remember if there was a specific reason, although this is just for building packages, so does it matter?

Habbie: Oh right, this is not the wforce Docker image. Ignore me :)

Habbie (approved):

PR #324, opened 2021-02-09
Support dynamic management of siblings and per-sibling encryption keys

neilcook:
This PR addresses two issues with the current version of wforce:
1) the single encryption key that has to be shared between all wforce instances
2) the inability to dynamically add siblings, which would extremely useful in environments such as kubernetes, where cluster membership could be highly dynamic

Utility functions for adding/removing/setting siblings are also added, which are used to manage siblings from Lua as well as from the new REST API commands to manage siblings.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -95,6 +95,7 @@ stages:
   - name: deploy
     if: branch = master AND env(DOCKER_USERNAME) IS present AND env(DOCKER_PASSWORD) IS present
 before_script:
+ - echo "$DOCKER_PASSWORD"
  - echo "$DOCKER_PASSWORD" | docker login --username "$DOCKER_USERNAME" --password-stdin
  - git clone https://github.com/jupp0r/prometheus-cpp.git
  - ( cd prometheus-cpp && git checkout tags/v0.9.0 -b v0.9.0 )
--- a/.travis.yml
+++ b/.travis.yml
@@ -95,7 +95,6 @@ stages:
   - name: deploy
     if: branch = master AND env(DOCKER_USERNAME) IS present AND env(DOCKER_PASSWORD) IS present
 before_script:
- - echo "$DOCKER_PASSWORD"
  - echo "$DOCKER_PASSWORD" | docker login --username "$DOCKER_USERNAME" --password-stdin
  - git clone https://github.com/jupp0r/prometheus-cpp.git
  - ( cd prometheus-cpp && git checkout tags/v0.9.0 -b v0.9.0 )
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,5 @@
 language: cpp
 dist: trusty
-env:
-  global:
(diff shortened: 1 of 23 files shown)


wforce/wforce-sibling.cc
```diff
@@ -179,3 +217,363 @@ void Sibling::queueMsg(const std::string& msg)
...
+    output_buffer += string("Unable to decode ") + key + " as Base64";
+    errlog("%s", output_buffer);
+    return false;
+  }
+
+  try {
+    parseSiblingString(address, ca, proto);
+  }
+  catch (const WforceException& e) {
+    const std::string errstr = (boost::format("%s [%s]. %s (%s)") % "addSibling() error parsing address/port" %
```
chbruyand, line 428:
```suggestion
    const std::string errstr = (boost::format("%s [%s]. %s (%s)") % "addSiblingWithKey() error parsing address/port" %
```

wforce/wforce-replication.cc
```diff
@@ -97,6 +104,11 @@ void replicateOperation(const ReplicationOperation& rep_op)
   encryptMsg(msg, packet);
 
   for(auto& s : *siblings) {
+    if (s->d_has_key) {
+      if (s->d_key != g_key) {
+        encryptMsgWithKey(msg, packet, s->d_key, s->d_nonce, s->mutx);
```
chbruyand, line 109: I must be missing something but I don't get why `msg` is added multiple times to `packet`

wforce/wforce-web.cc
```diff
@@ -211,41 +215,220 @@ void parseSyncCmd(const YaHTTP::Request& req, YaHTTP::Response& resp, const std:
...
+{
+  using namespace json11;
+  Json msg;
+  string err;
+
+  resp.status = 200;
+
+  msg = Json::parse(req.body, err);
+  if (msg.is_null()) {
+    resp.status = 500;
```
chbruyand, line 271: Maybe a 4xx response code would be more relevant ?

.travis.yml
```diff
@@ -85,6 +82,8 @@ stages:
   - name: deploy
     if: branch = master AND env(DOCKER_USERNAME) IS present AND env(DOCKER_PASSWORD) IS present
 before_script:
+ - echo "$DOCKER_PASSWORD"
```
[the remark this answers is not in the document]
neilcook: Oops, no!

common/sstuff.hh
```diff
@@ -180,6 +181,45 @@ public:
...
+    arg |= O_NONBLOCK;
+    if(fcntl(d_socket, F_SETFL, arg) < 0) {
+      throw NetworkError(strerror(errno));
+    }
+    if(::connect(d_socket,(struct sockaddr *)&ep, ep.getSocklen()) < 0) {
+      if (errno != EINPROGRESS) {
+        throw NetworkError(strerror(errno));
+      }
+      else {
+        struct pollfd pf = { d_socket, POLLWRNORM, 0 };
```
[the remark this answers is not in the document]
neilcook: Well I want to check I can write, so POLLOUT (which is equivalent to POLLWRNORM) rather than POLLIN?

wforce/wforce-replication.cc
neilcook: Yes, this is a bug. It should use the original packet if there is no unique key for the sibling, otherwise it should encrypt msg into a different packet variable.

wforce/wforce-web.cc
neilcook: Yes agreed.

common/sstuff.hh
```diff
@@ -180,6 +181,45 @@ public:
...
+        int saved_errno = errno;
+        if((arg = fcntl(d_socket, F_GETFL, NULL)) < 0) {
+          throw NetworkError(strerror(errno));
+        }
+        arg &= (~O_NONBLOCK);
+        if(fcntl(d_socket, F_SETFL, arg) < 0) {
+          throw NetworkError(strerror(errno));
+        }
+        if (poll_ret <=0) {
+          throw NetworkError(strerror(saved_errno));
```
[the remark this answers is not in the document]
neilcook: yes good point

wforce/wforce-web.cc
neilcook: Yes, will fix throughout this file

common/sstuff.hh
```diff
@@ -180,6 +181,45 @@ public:
       throw NetworkError(strerror(errno));
   }
 
+  //! Connecto the socket to a specific endpoint with a configurable timeout in milliseconds
+  void connectWithTimeout(const ComboAddress &ep, int timeout)
```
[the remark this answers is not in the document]
neilcook: FYI I've decided to keep rather than rewrite to use SConnectWithTimeout, as that doesn't do any of the non-blocking setting/unsetting.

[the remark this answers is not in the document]
chbruyand: OK. FYI there are `set{Non,}Blocking()` functions in `common/misc.hh`

chbruyand:
Apart from tests that fail, LGTM!

PR #337, opened 2021-05-05
Fix issues reported by `cppcheck`

chbruyand:


common/device_parser.cc
```diff
@@ -107,9 +107,9 @@ bool DeviceCache::readFromCache(const std::string& device_id, std::map<std::stri
   return false;
 }
 
-void DeviceCache::addToCache(const std::string& device_id, const std::map<std::string, std::string> device_attrs)
+void DeviceCache::addToCache(const std::string& device_id, std::map<std::string, std::string>&& device_attrs)
 {
   WriteLock wl(&d_rwlock);
 
-  d_devicemap.emplace(std::make_pair(device_id, std::move(device_attrs)));
+  d_devicemap.emplace(device_id, std::move(device_attrs));
```
neilcook, line 114: I think the std::move here is redundant since it's already an rvalue?

chbruyand: Yes of course!

chbruyand: Hum no, `device_attrs` is considered as a lvalue in the current expression thus not moved by default.

neilcook: But map::emplace is defined as:
`template <class... Args>  pair<iterator,bool> emplace (Args&&... args);`

So this shouldn't this "just work" if device_attrs is already a universal reference?

neilcook: I just looked it up and I'm wrong :)

I wonder if we should use std::forward rather than move here? Then the caller is in control

neilcook: In fact I'm positive we should use std::forward here, see my other comment

common/login_tuple.cc
```diff
@@ -148,7 +148,7 @@ void LoginTuple::setDeviceAttrs(const json11::Json& msg, const std::shared_ptr<U
       device_attrs.insert(std::make_pair("app.minor", oxmad.app.minor));
       device_attrs.insert(std::make_pair("device.family", oxmad.device.family));
     }
-    dcache.addToCache(my_device_id, device_attrs);
+    dcache.addToCache(my_device_id, std::move(device_attrs));
```
neilcook, line 151: We can't use std::move here - device_attrs is a member variable and this will leave it undefined, and it's needed. That's why the test is failing

common/device_parser.cc
neilcook: The other option is simply to get rid of all the move stuff, because I'm just adding to a cache, and there's absolutely no reason not to copy it

neilcook (approved):

neilcook:
LGTM!

PR #338, opened 2021-05-06
Refactor wforce-replication.cc from a collection of functions into a class

neilcook:
This not only provides better modularity, but it also allows key
functions to be overidden in subclasses. This change affects quite a lot
of other files, so this is a fairly chunky commit.
Using a class also replaces a whole bunch of globals with a single global for
the class instance.


chbruyand (approved):

neilcook:
thanks Max!

PR #339, opened 2021-05-12
Various fixes

neilcook:
- Update builder submodule
- Make the WforceReplication class more usable by subclasses
- Remove reference to a global in WforceReplication class
- Update spec file to use list tag in Release to match pdns standards
- Fix a coverity-reported issue

### DIFF
--- a/trackalert/trackalert-lua.cc
+++ b/trackalert/trackalert-lua.cc
@@ -298,6 +298,21 @@ vector<std::function<void(void)>> setupLua(bool client, bool multi_lua,
   }
 
   if (!multi_lua) {
+    c_lua.writeFunction("setKey", [](const std::string& key) {
+      string newkey;
+      if(B64Decode(key, newkey) < 0) {
+        g_outputBuffer=string("Unable to decode ")+key+" as Base64";
+        errlog("%s", g_outputBuffer);
+      }
+      else
+        g_key = newkey;
+    });
+  }
+  else {
+    c_lua.writeFunction("setKey", [](const std::string& key) { });
+  }
+
+  if (!multi_lua) {
     c_lua.writeFunction("testCrypto", [](string testmsg)
     {
       try {
--- a/trackalert/trackalert-lua.cc
+++ b/trackalert/trackalert-lua.cc
@@ -298,14 +298,17 @@ vector<std::function<void(void)>> setupLua(bool client, bool multi_lua,
   }
 
   if (!multi_lua) {
-    c_lua.writeFunction("setKey", [](const std::string& key) {
+    c_lua.writeFunction("setKey", [](const std::string& key) -> bool {
       string newkey;
       if(B64Decode(key, newkey) < 0) {
         g_outputBuffer=string("Unable to decode ")+key+" as Base64";
         errlog("%s", g_outputBuffer);
+        return false;
       }
-      else
+      else {
         g_key = newkey;
+        return true;
+      }
     });
   }
   else {
--- a/wforce/wforce-lua.cc
+++ b/wforce/wforce-lua.cc
@@ -1003,20 +1003,36 @@ vector<std::function<void(void)>> setupLua(bool client, bool multi_lua, LuaConte
   }
 
   if (!multi_lua) {
+    c_lua.writeFunction("setKey", [](const std::string& key) {
+      string newkey;
+      if(B64Decode(key, newkey) < 0) {
+        g_outputBuffer=string("Unable to decode ")+key+" as Base64";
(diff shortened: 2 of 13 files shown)


wforce/wforce-replication.hh
```diff
@@ -46,18 +46,20 @@ public:
   void setNumSiblingThreads(unsigned int num_threads) { d_num_sibling_threads = num_threads; }
   GlobalStateHolder<vector<shared_ptr<Sibling>>>& getSiblings() { return d_siblings; }
   void replicateOperation(const ReplicationOperation& rep_op);
+  void setEncryptionKey(const std::string& key) { d_key = key; }
+  std::string getEncryptionKey() { return d_key; }
```
omoerbeek, line 50:
```suggestion
  std::string getEncryptionKey() const { return d_key; }
```

wforce/wforce-lua.cc
```diff
@@ -1002,21 +1002,37 @@ vector<std::function<void(void)>> setupLua(bool client, bool multi_lua, LuaConte
     c_lua.writeFunction("showVersion", []() { });
   }
 
+  if (!multi_lua) {
+    c_lua.writeFunction("setKey", [](const std::string& key) {
```
omoerbeek, line 1006: I'd prefer if this function returned a status, indicating success/failure.

wforce/wforce-replication.hh
```diff
@@ -46,18 +46,20 @@ public:
...
 protected:
-  bool checkConnFromSibling(const ComboAddress& remote, shared_ptr<Sibling>& recv_sibling);
+  virtual bool checkConnFromSibling(const ComboAddress& remote, shared_ptr<Sibling>& recv_sibling);
   void parseTCPReplication(std::shared_ptr<Socket> sockp, const ComboAddress& remote, std::shared_ptr<Sibling> recv_sibling);
   void parseReceivedReplicationMsg(const std::string& msg, const ComboAddress& remote, std::shared_ptr<Sibling> recv_sibling);
-private:
```
omoerbeek, line 53: Was the removal of `private:` intended?

trackalert/trackalert-lua.cc
```diff
@@ -297,6 +297,21 @@ vector<std::function<void(void)>> setupLua(bool client, bool multi_lua,
     c_lua.writeFunction("showVersion", []() { });
   }
 
+  if (!multi_lua) {
```
omoerbeek, line 300: A bit confusing, `g_key` was being abandoned but is still used here???

omoerbeek (commented):
Are the builder changes intended to be included?
Lacking the big picture, a few nits and a question.

wforce/wforce-replication.hh
neilcook: yes it was - these all need to be accessed from a subclass.

wforce/wforce-replication.hh
```diff
@@ -46,18 +46,20 @@ public:
   void setNumSiblingThreads(unsigned int num_threads) { d_num_sibling_threads = num_threads; }
   GlobalStateHolder<vector<shared_ptr<Sibling>>>& getSiblings() { return d_siblings; }
   void replicateOperation(const ReplicationOperation& rep_op);
+  void setEncryptionKey(const std::string& key) { d_key = key; }
+  std::string getEncryptionKey() { return d_key; }
 protected:
-  bool checkConnFromSibling(const ComboAddress& remote, shared_ptr<Sibling>& recv_sibling);
+  virtual bool checkConnFromSibling(const ComboAddress& remote, shared_ptr<Sibling>& recv_sibling);
```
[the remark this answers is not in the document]
neilcook: Yes indeed they probably all should be - I only made virtual the ones I want to override in a specific subclass, but that is is somewhat blinkered

trackalert/trackalert-lua.cc
neilcook: That's because the trackalert daemon doesn't use the WforceReplication class - it only uses encryption for the console, hence still using g_key.

wforce/wforce-replication.hh
neilcook: > Are the builder changes intended to be included?
> Lacking the big picture, a few nits and a question.

The big picture is that weakforced is used as a submodule in another project, so the whole WforceReplication class was created to reduce code duplication in that project

wforce/wforce-lua.cc
neilcook: sure

neilcook:
The builder changes are intentional in so far as I like to keep it up to date

Issue #345, opened 2021-07-21
[BUG] GeoIP2 is not build without legacy GeoIP lib


slavkoja:
**Describe the bug**
Have GeoIP2 libraries installed is not enough to build-in its support and the legacy GeoIP library is required.

**To Reproduce**
I use configure from debian package, where i setup:
```
        --with-luajit \
        --sysconfdir=/etc/wforce \
        --enable-trackalert
```
Which results in configure called with:
```
./configure --build=x86_64-linux-gnu --prefix=/usr --includedir=\${prefix}/include --mandir=\${prefix}/share/man --infodir=\${prefix}/share/info --sysconfdir=/etc --localstatedir=/var --disable-option-checking --disable-silent-rules --libdir=\${prefix}/lib/x86_64-linux-gnu --runstatedir=/run --disable-maintainer-mode --disable-dependency-tracking --with-luajit --sysconfdir=/etc/wforce --enable-trackalert
```

With installed only libmaxminddb library, the confgure script outputs:
```
checking for geoip... no
checking for libmaxminddb... yes
```
Final binary is not linked against libmaxminddb:
```
ldd /usr/bin/wforce | grep -iP "geoip|maxm"
<return nothing>
```
Of course, GeoIP2 fails even to initialize DB.

With installed both, libgeoip and libmaxminddb, the configure script outputs:
```
checking for geoip... yes
checking for libmaxminddb... yes
```
Final binary shows:
```
ldd /usr/bin/wforce | grep -iP "geoip|maxm"
	libGeoIP.so.1 => /lib/x86_64-linux-gnu/libGeoIP.so.1 (0x00007fb54a020000)
	libmaxminddb.so.0 => /lib/x86_64-linux-gnu/libmaxminddb.so.0 (0x00007fb54a018000)
```
And GeoIP2 initialize and lookup works:
```
Read configuration from '/etc/wforce/wforce.conf'
Opened MMDB database /usr/local/GeoLite2-Country.mmdb (type: GeoLite2-Country version: 2.0)
```

**Expected behavior**

I would expect to be able to build GeoIP2 support without needing legacy libraries.

**OS (please complete the following information):**
 - OS: Debian
 - Version: testing, bullseye

neilcook:
I'd need to see more information to reproduce this. Please upload your config.log and the build output. On my Mac, the exact setup you describe above (no legacy geoip, but libmaxminddb installed) works fine:

```
[...]

Issue #344, opened 2021-07-19
[BUG]


slavkoja:
**Describe the bug**
I tried to build weakforced package for debian testing (bullseye) and it fails to link final `wforce` binary. The linking fails with:
```
/bin/bash ../libtool  --tag=CXX   --mode=link g++  -g -O2 -ffile-prefix-map=/build/weakforced-2.4.1=. -fstack-protector-strong -Wformat -Werror=format-security   -Wl,-z,relro -o wf_dump_entries dump_entries.o /build/weakforced-2.4.1/common/libweakforce.la -lcurl -lsodium 
libtool: link: g++ -g -O2 -ffile-prefix-map=/build/weakforced-2.4.1=. -fstack-protector-strong -Wformat -Werror=format-security -Wl,-z -Wl,relro -o wf_dump_entries dump_entries.o  /build/weakforced-2.4.1/common/.libs/libweakforce.a -L/build/weakforced-2.4.1/ext/yahttp/yahttp -L/build/weakforced-2.4.1/ext/json11 -lboost_filesystem -lboost_system -lmaxminddb -lprometheus-cpp-core -lcurl -lsodium
/bin/bash ../libtool  --tag=CXX   --mode=link g++  -g -O2 -ffile-prefix-map=/build/weakforced-2.4.1=. -fstack-protector-strong -Wformat -Werror=format-security   -Wl,-z,relro -o wforce wforce.o wforce-lua.o wforce-web.o wforce-sibling.o twmap.o twmap-wrapper.o blackwhitelist.o replication.o replication_sdb.o replication_bl.o replication_wl.o wforce-replication.o wforce-prometheus.o replication.pb.o -lreadline -lluajit-5.1 -L/build/weakforced-2.4.1/ext/yahttp/yahttp -lyahttp /build/weakforced-2.4.1/ext/ext/libext.la /build/weakforced-2.4.1/common/libweakforce.la -lsodium  -lmaxminddb -lgetdns  -lprotobuf -lsystemd -L/build/weakforced-2.4.1/ext/json11 -ljson11 -lboost_date_time -lboost_regex -lhiredis -lcurl  -lcrypto -lyaml-cpp -L/usr/lib -lprometheus-cpp-core 
[...]

PR #347, opened 2021-08-05
Various packaging and compilation fixes

neilcook:
- Fix pthread link issue on debian bullseye (#344)
- Ensure geoip2 function inclusion is not dependent on geoip legacy lib availability (#345)
- Tidy up debian control files so that they work for debian stretch through bullseye
- Add packaging target for debian bullseye


neilcook:
thanks habbie!

Issue #355, opened 2021-12-07
[BUG] wforce logs errors about failing to create folders ./uploads/tmp/XX


mgehlmann:
**Describe the bug**
When I start up wforce v2.6.0 (used builder script) on Debian bullseye with default config, I get these errors in the log:
```
Dec  7 09:20:29 wforce0 wforce[250278]: 20211207 08:20:29.901676 UTC 250290 ERROR Can't create path:./uploads/tmp/00 - Utilities.cc:1098
Dec  7 09:20:29 wforce0 wforce[250278]: 20211207 08:20:29.901950 UTC 250290 ERROR Can't create path:./uploads/tmp/01 - Utilities.cc:1098
[...]
Dec  7 09:20:29 wforce0 wforce[250278]: 20211207 08:20:29.917911 UTC 250290 ERROR Can't create path:./uploads/tmp/FF - Utilities.cc:1098
```
I did not find any information about the `./uploads` in the docs, though.
I tested a bit and on one occasion, which I cannot reproduce, the folders 00-03 were created in `/etc/wforce/uploads/tmp/`:

```
# tree /etc/wforce/
/etc/wforce/
├── regexes.yaml
├── uploads
│   └── tmp
│       ├── 00
│       ├── 01
│       └── 02
└── wforce.conf
```

This time there are only error messages for folders 03-FF. Additional restarts do not lead to more created folders.

Could the issue be, that `/etc/wforce` appears to be the working directory of the process? The systemd service file that is installed by the package sets:
```
User=wforce
Group=wforce
```
But `/etc/wforce` is created with 
```
# ls -ld /etc/wforce
drwxr-xr-x 2 root root 4096  7. Dez 10:57 /etc/wforce
```
So the creation would probably fail due to missing permissions.
The working directory should probably be somewhere in `/var/`?

**To Reproduce**
1. Build wforce 2.6.0 for debian-bullseye using builder script
2. install .deb package
3. start service with default config

**Expected behavior**
Not sure, since I don't know what `./uploads` is for, but I suppose that all folders 00-FF should be created, and probably not in `/etc/wforce`, but somewhere in `/var`.

**OS (please complete the following information):**
 - OS: Debian
 - bullseye

neilcook:
[...]

PR #356, opened 2021-12-07
Fix upload warnings on startup and syncDone issue

neilcook:
The drogon HTTP framework we now use for the REST API also allows file uploads, which we don't use. Unfortunately, by default it tries to create a bunch of directories in the working directory, which is usually not writable (e.g. /etc/wforce), as described in #355.

This PR modifies the drogon upload directory so that the warnings do not occur.

This PR also ensures that before the syncDB functions are called, that the webserver is running, otherwise it sleeps until the webserver is running. This fixes the issue described in #358.

### DIFF
--- a/common/wforce-webserver.hh
+++ b/common/wforce-webserver.hh
@@ -147,7 +147,7 @@ public:
       drogon::app().disableSigtermHandling();
       drogon::app().setLogLevel(trantor::Logger::kWarn);
       // We will never allow uploads, but drogon wants to create a bunch of temp files in uploadPath/tmp/xx
-      drogon::app().setUploadPath("/var");
+      drogon::app().setUploadPath("/tmp/wforce");
       // Set custom 404 response
       auto resp = drogon::HttpResponse::newHttpResponse();
       resp->setBody(R"({"status":"failure", "reason":"Not found"})");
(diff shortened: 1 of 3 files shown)


common/wforce-webserver.hh
```diff
@@ -146,6 +146,8 @@ public:
       drogon::app().disableSession();
       drogon::app().disableSigtermHandling();
       drogon::app().setLogLevel(trantor::Logger::kWarn);
+      // We will never allow uploads, but drogon wants to create a bunch of temp files in uploadPath/tmp/xx
+      drogon::app().setUploadPath("/tmp/wforce");
```
chbruyand, line 150: Maybe use `std::filesystem::temp_directory_path` ?

chbruyand (approved):

common/wforce-webserver.hh
neilcook: Ok, not such a good idea as it turn out, not for C++ reasons but for toolchain reasons:
```
/usr/bin/ld: wforce-lua.o: in function `WforceWebserver::init()':
3779
/wforce/wforce/../common/wforce-webserver.hh:151: undefined reference to `std::filesystem::temp_directory_path[abi:cxx11]()'
```

Support for std::filesystem is a something of a toolchain nightmare, and one I don't want to get into for this patch. So I'm going to revert this change for now.

PR #357, opened 2021-12-07
Use debian-bullseye-slim for docker image saving over 100MB and also

neilcook:
- Use debian bullseye-slim to save space (over 100MB)
- Check for users providing their own config file (e.g. via a volume mount) and use that instead of generating a new config file

### DIFF
--- a/docker/wforce_image/docker-entrypoint.sh
+++ b/docker/wforce_image/docker-entrypoint.sh
@@ -18,6 +18,11 @@ if [[ "x$TRACKALERT" == "x" ]]; then
     if [[ "$WFORCE_VERBOSE" -eq "1" ]]; then
         WFORCE_CMD="$WFORCE_CMD -v"
     fi
+    # Check if they just supplied their own wforce.conf file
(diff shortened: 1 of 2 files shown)


docker/wforce_image/docker-entrypoint.sh
```diff
@@ -18,6 +18,11 @@ if [[ "x$TRACKALERT" == "x" ]]; then
     if [[ "$WFORCE_VERBOSE" -eq "1" ]]; then
         WFORCE_CMD="$WFORCE_CMD -v"
     fi
+    # Check if they just supplied their own wforce.conf file
+    if [[ -f /etc/wforce/wforce.conf ]]; then
+      echo "Using supplied wforce config file"
+      WFORCE_CONFIG_FILE=/etc/wforce/wforce.conf
```
chbruyand, line 24:
```suggestion
      echo "Using supplied wforce config file /etc/wforce/wforce.conf"
      WFORCE_CONFIG_FILE="/etc/wforce/wforce.conf"
```

chbruyand (approved):

PR #359, opened 2021-12-20
Fix crash bug when scheduler runs before multi-lua states are created

neilcook:


trackalert/trackalert.cc
```diff
@@ -634,12 +634,13 @@ try
 
   // Initialise Prometheus Metrics
   initPrometheusMetrics(std::make_shared<PrometheusMetrics>("trackalert"));
-  
-  // this sets up the global lua state used for config and setup
-  auto todo=setupLua(false, false, g_lua, g_report, nullptr, g_custom_func_map, g_cmdLine.config);
 
-  // now we setup the allow/report lua states
+  // now we setup the multi-lua lua states (we do this first because there are routines such as the scheduler
+  // in the global config that use the multi-lua states
```
Habbie, line 639: missing `)`

Habbie (changes requested):
missing char in a comment, otherwise LGTM!

trackalert/trackalert.cc
neilcook: sharp eyes - thanks!

PR #364, opened 2022-03-10
Enable ip and login substitution in blacklist return messages

neilcook:
Example template: "your IP address {ip} has been blocked temporarily for bad behaviour"
This is so that clients can get messages such as:
"Your IP address 8fea::1 has been blocked temporarily for bad behaviour"

Similarly the the template: "Your login {login} for IP address {ip} has been blocked due to excessive badness"
etc.

This PR does mean that in all cases a blacklist response will incur an extra allocation for a new string compared with before, even if there are no {ip} or {login} templates in the message.

### DIFF
--- a/wforce/wforce-web.cc
+++ b/wforce/wforce-web.cc
@@ -41,6 +41,7 @@
 #include "webhook.hh"
 #include "boost/date_time/posix_time/posix_time.hpp"
 #include "boost/date_time/gregorian/gregorian.hpp"
+#include "boost/algorithm/string.hpp"
 #include "boost/regex.hpp"
 #include "wforce-prometheus.hh"
 #include "wforce-sibling.hh"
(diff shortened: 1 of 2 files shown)


wforce/wforce-web.cc
```diff
@@ -772,6 +773,33 @@ enum AllowReturnFields {
...
+  std::string new_msg;
+
+  if (msg.find(ipPattern) != std::string::npos)
+    replace_ip = true;
+  if (msg.find(loginPattern) != std::string::npos)
+    replace_login = true;
+
+  if (replace_ip || replace_login)
+    // The maximum length of an IPv6 address string is 39 characters
+    new_msg.reserve(msg.length()+39+login.length()); // This should prevent a realloc in the majority of cases
```
Habbie, line 792: this reserves for one instance of the IP

wforce/wforce-web.cc
```diff
@@ -772,6 +773,33 @@ enum AllowReturnFields {
...
+
+  if (replace_ip)
+    new_msg = boost::algorithm::replace_all_copy(msg, "{ip}", ca.toString());
```
Habbie, line 795: but this might put in several

Habbie (approved):

PR #368, opened 2022-04-13
Logging improvements

neilcook:
Previous to this PR, stdout would log everything, even debug logs. Additionally, the new webserver would log only errors, with no way to debug issues.

This PR:
- Adds a log level command line flag to wforce and trackalert
- Adds a global log level variable used by dolog.hh used to filter logging to stdout
- Uses the configured log level to set drogon webserver log level (which also logs to stdout - it has no syslog capability)

### DIFF
--- a/common/dolog.hh
+++ b/common/dolog.hh
@@ -74,6 +74,10 @@ extern bool g_console;
 extern bool g_verbose;
 extern bool g_docker;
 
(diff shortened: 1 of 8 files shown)


Habbie (approved):

neilcook:
thanks Peter!

PR #369, opened 2022-05-05
Warmup

mrannanj:


neilcook:
You should rebase this against master, as I fixed some GitHub actions issues

regression-tests/test_LogLevel.py
```diff
@@ -0,0 +1,28 @@
...
+        assert content in out, out
+
+    def test_debug(self):
+        self.check_loglevel(7, "TRACE")
+
+    def test_info(self):
+        self.check_loglevel(5, "ACL allowing queries from")
+
+    def test_warning(self):
+        self.check_loglevel(4, "Warning - webserver() configuration")
```
neilcook, line 28: This is probably not a good check - at some point the config will move to using the non-deprecated addListener() config, at which point this log line will disappear and this test will randomly fail.

neilcook (approved):
lgtm

PR #373, opened 2022-07-04
Various fixes inspired by work on deploying weakforced in K8S

neilcook:
- Fix an issue where IPv6 ComboAddress returned zero port number (which caused v6 HTTP listen addresses to not work) 
- Update to use more recent cmake, prometheus-cpp and libdrogon releases
- Fix build issues caused by the updates to prometheus and drogon
- Set V6ONLY socket option to stop v6 sockets from managing v4 addresses for replication
- Push an unstable docker image tag to docker-hub from the master branch
- Add net-tools to the wforce docker image image for easier debugging
- Add config option to disable password for /metrics endpoint


Habbie (approved):
LGTM. Some of the build instructions are repeated in a lot of files though!

neilcook:
Yes I need to  to rationalise this, as the same (or similar) build instructions are split between different dockerfiles (regression, builder), GitHub Actions etc. It's a bit of a mess... another PR methinks...

PR #376, opened 2022-08-02
Fix wforce docker image issue

neilcook:
- Image was using the sample wforce and trackalerrt config files instead of "no"config file as should be the case
- Generate the console/replication key from kernel random uuid rather than try to use wforce itself to generate it (which can lead to a deadlock)
- Don't use deprecated version of webserver listen command when generating config
- Test wforce image properly in regression (previously some tests were failing but not being picked up)
- Test trackalert binary in regression

### DIFF
--- a/docker/wforce_image/wforce.conf.j2
+++ b/docker/wforce_image/wforce.conf.j2
@@ -1,7 +1,7 @@
 -- This configuration file is provided as an example
 -- It only scratches the surface of what weakforced policy can do
 
-addListener("0.0.0.0:{{env['WFORCE_HTTP_PORT']}}", false,    "",             "",         {})
+addListener("0.0.0.0:{{env['WFORCE_HTTP_PORT']}}", false, "", "", {})
 setWebserverPassword("{{env['WFORCE_HTTP_PASSWORD']}}")
 
 {{env['WFORCE_KEY']}}
--- a/docker/wforce_image/wforce.conf.j2
+++ b/docker/wforce_image/wforce.conf.j2
@@ -1,7 +1,9 @@
 -- This configuration file is provided as an example
 -- It only scratches the surface of what weakforced policy can do
 
-webserver("0.0.0.0:{{env['WFORCE_HTTP_PORT']}}", "{{env['WFORCE_HTTP_PASSWORD']}}")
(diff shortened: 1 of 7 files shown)


docker/wforce_image/wforce.conf.j2
```diff
@@ -1,7 +1,9 @@
 -- This configuration file is provided as an example
 -- It only scratches the surface of what weakforced policy can do
 
-webserver("0.0.0.0:{{env['WFORCE_HTTP_PORT']}}", "{{env['WFORCE_HTTP_PASSWORD']}}")
+addListener("0.0.0.0:{{env['WFORCE_HTTP_PORT']}}", false,    "",             "",         {})
```
Habbie, line 4: all this whitespace seems pointless without a comment like in `trackalert.conf`

trackalert/trackalert.conf
```diff
@@ -1,4 +1,10 @@
-webserver("0.0.0.0:8085", "--WEBPWD")
+-- Old way to configure the webserver
+-- webserver("0.0.0.0:8085", "--WEBPWD")
+-- New way to configure the webserver
+--          IP addr:port    Use SSL? Certificate File Private Key TLS Options - see https://www.openssl.org/docs/manmaster/man3/SSL_CONF_cmd.html
+addListener("0.0.0.0:8085", false,    "",             "",         {})
+setWebserverPassword("--WEBPWD")
```
Habbie, line 6: is that password what you wanted to use here?

Habbie (commented):
two comments, do with them what you like. Approved in any case :)

trackalert/trackalert.conf
neilcook: No, good catch, thanks!

docker/wforce_image/wforce.conf.j2
neilcook: Yes, indeed I will add the additional comments as per `trackalert.conf`

trackalert/trackalert.conf
neilcook: Oh, no this is trackalert.conf not trackers.conf.j2 - this is what I want here!

PR #381, opened 2022-11-11
Add missing dependencies to spec file

fraenki:
While trying to build a RPM package I've noticed that several dependencies are not mentioned in the spec file, so I've added them:

```
+BuildRequires: drogon
+BuildRequires: prometheus-cpp-devel
+BuildRequires: yaml-cpp-devel
```

Besides that I've sorted the list of dependencies alphanumerically to make it easier to grasp :)

### DIFF
--- a/builder-support/specs/wforce.spec
+++ b/builder-support/specs/wforce.spec
@@ -16,7 +16,6 @@
 %define __prelink_undo_cmd %{nil}
 %global __os_install_post %(echo '%{__os_install_post}' | sed -e 's!/usr/lib[^[:space:]]*/brp-python-bytecompile[[:space:]].*$!!g')
 
-
 %global wforce_restart_flag /var/run/wforce-restart-after-rpm-install
 %global trackalert_restart_flag /var/run/wforce-trackalert-restart-after-rpm-install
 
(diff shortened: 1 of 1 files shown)


neilcook:
`BuildRequires: yaml-cpp-devel`

This makes sense.

```
BuildRequires: drogon
BuildRequires: prometheus-cpp-devel
```

This I do not understand, since these packages are not available, at least on centos 7 or oracle linux 8.

fraenki:
> This I do not understand, since these packages are not available, at least on centos 7 or oracle linux 8.

That's true (I've created source RPMs for them). However, they are required, otherwise the build would fail. Tested with weakforced 2.6.2. I have no insights into the weakforced build environment, so I don't know how this problem is solved there.

neilcook:
> > This I do not understand, since these packages are not available, at least on centos 7 or oracle linux 8.
> 
> That's true (I've created [source] RPMs for them). However, they are required, otherwise the build would fail. Tested with weakforced 2.6.2. I have no insights into the weakforced build environment, so I don't know how this problem is solved there.

So your changes would in fact break the weakforced build environment, as you can see by the failed CI. 

I highly recommend using the weakforced build environment to build rpms; it's extremely simple to use:
```
git submodule init
builder/build.sh centos-7
builder/build.sh ol-8
```
etc.

You can add new targets by adding the appropriate docker files in builder-support/dockerfiles.

fraenki:
OK, I see that you're installing the dependencies from source in builder:

https://github.com/PowerDNS/weakforced/blob/2542d1f3a8979663648141452f0175b9d65db998/builder-support/dockerfiles/Dockerfile.rpmbuild#L34-L44 

We won't use the builder, though, because we'd like to build all RPMs in the same way, hence the suggested additions to the SPEC file. I understand that it's not useful to weakforced, so I'll close this PR.

PR #388, opened 2023-02-07
Build Libmaxminddb ourselves and python setup tools version fix

neilcook:


Habbie:
Remind me, why not grab it from EPEL archives?

Habbie:
> Remind me, why not grab it from EPEL archives?

Ah, it's just two lines, and it's in Docker - and then users don't have to bother with EPEL archives. Makes sense!

Habbie (approved):

Habbie:
> As Neil told me yesterday, it's because Amazon Linux 2 does not have access to it anymore (since it is no longer in EPEL, and it is not part of AL2 core). So this is a requirement to run on AL2, not an optimization.

This is clear to me - I even spoke to some Fedora staff to see if they would bring it back (they wouldn't). At the time I suggested Neil could `wget` the last version from the EPEL archives, which works, but does not help users of our packages, which is why the static build in here indeed makes sense.

PR #392, opened 2023-02-24
Use trackalert  to create SETKEY

bgmex:
Minor fix

### DIFF
--- a/builder-support/debian/wforce-trackalert.postinst
+++ b/builder-support/debian/wforce-trackalert.postinst
@@ -17,7 +17,7 @@ case "$1" in
       echo "done"
     fi
     echo -n "Modifying trackalert.conf to replace password and key..."
-    SETKEY=`echo "makeKey()" | wforce -c | grep setKey`
+    SETKEY=`echo "makeKey()" | trackalert | grep setKey`
     WEBPWD=`dd if=/dev/urandom bs=1 count=32 2>/dev/null | base64 | rev | cut -b 2-14 | rev`
     sed -e "s#--WEBPWD#$WEBPWD#" -e "s#--SETKEY#$SETKEY#" -i $TRACKALERTCONF
     echo "done"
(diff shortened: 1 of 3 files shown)


neilcook:
So, the -c flag makes trackalert run as a client, connecting to a server. This will not generate a key, it will simply fail to connect with an error message. I'm not sure what the reason for this PR is? Is the current post install script not working?

bgmex:
I'm sorry for my poor PR. I forgot to add the explanation.

Meanwhile I did some further research and learned, that my PR is insufficient/unnecessary.

The current post install script for wforce-trackalert only fails if wforce is already running. My PR fixes this.
If wforce is not running my PR fails and the current post install scripts works.

A proper fix would be something like:
```
systemctl is-active --quiet wforce
if [ $? -eq 0 ]; then
    # wforce is running so connect to it using -c
    SETKEY=`echo "makeKey()" | wforce -c | grep setKey`
else
    # wforce is not running
    SETKEY=`echo "makeKey()" | wforce | grep setKey`
fi
```
Or we that for granted, that wforce is not running when wforce-trackalert is being installed

neilcook:
Actually the proper fix is to run trackalert to generate the key not wforce, as trackalert should have no dependency on wforce being installed:
```
SETKEY=`echo "makeKey()" | trackalert | grep setKey`

**Comment by neilcook** [comment]
Note that this issue also exists in the spec file for RH packages

File: builder-support/debian/wforce-trackalert.postinst [hunk 1]
```diff
@@ -17,7 +17,7 @@ case "$1" in
       echo "done"
     fi
     echo -n "Modifying trackalert.conf to replace password and key..."
-    SETKEY=`echo "makeKey()" | wforce | grep setKey`
+    SETKEY=`echo "makeKey()" | wforce -c | grep setKey`
```
neilcook, line 20:
```suggestion
    SETKEY=`echo "makeKey()" | trackalert -c | grep setKey`
```

bgmex: ```trackalert -c``` needs trackalert.service to be running. While installing trackalert we can assume it's not running
```suggestion
    SETKEY=`echo "makeKey()" | trackalert | grep setKey`
```

neilcook: yes this change would work

neilcook (approved):

PR #399, opened 2023-07-11
Use OpenSSL crypto instead of libsodium when libsodium isn't available

neilcook:
- OpenSSL version of sodencrypt.cc/hh
- Regression now includes two versions - one testing with libsodium, one testing without (using this new code)

### DIFF
--- a/common/sodcrypto.cc
+++ b/common/sodcrypto.cc
@@ -84,7 +84,7 @@ std::string sodEncryptSym(const std::string& msg, const std::string& key, Sodium
   int len;
   int ciphertext_len;
   // Each thread gets its own cipher context
-  thread_local auto ctx = std::unique_ptr<EVP_CIPHER_CTX, decltype(&EVP_CIPHER_CTX_free)>(nullptr, EVP_CIPHER_CTX_free);;
+  static thread_local auto ctx = std::unique_ptr<EVP_CIPHER_CTX, decltype(&EVP_CIPHER_CTX_free)>(nullptr, EVP_CIPHER_CTX_free);
 
   if (key.length() != CHACHA20_POLY1305_KEY_SIZE) {
     ostringstream err;
@@ -126,7 +126,7 @@ std::string sodDecryptSym(const std::string& msg, const std::string& key, Sodium
   int len;
   int plaintext_len;
   // Each thread gets its own cipher context
-  thread_local auto ctx = std::unique_ptr<EVP_CIPHER_CTX, decltype(&EVP_CIPHER_CTX_free)>(nullptr, EVP_CIPHER_CTX_free);;
+  static thread_local auto ctx = std::unique_ptr<EVP_CIPHER_CTX, decltype(&EVP_CIPHER_CTX_free)>(nullptr, EVP_CIPHER_CTX_free);
 
   if (key.length() != CHACHA20_POLY1305_KEY_SIZE) {
     ostringstream err;
--- a/common/sodcrypto.cc
+++ b/common/sodcrypto.cc
@@ -70,7 +70,7 @@ std::string sodDecryptSym(const std::string& msg, const std::string& key, Sodium
 string newKey()
 {
   unsigned char key[CHACHA20_POLY1305_KEY_SIZE];
-  if (!RAND_priv_bytes(key, sizeof key)) {
+  if (RAND_priv_bytes(key, sizeof key) != 1) {
     throw std::runtime_error(
         "Could not initialize random number generator for cryptographic functions - this is not recoverable");
   }
--- a/common/sodcrypto.cc
+++ b/common/sodcrypto.cc
@@ -84,7 +84,7 @@ std::string sodEncryptSym(const std::string& msg, const std::string& key, Sodium
   int len;
   int ciphertext_len;
   // Each thread gets its own cipher context
-  thread_local EVP_CIPHER_CTX *ctx = nullptr;
+  thread_local auto ctx = std::unique_ptr<EVP_CIPHER_CTX, decltype(&EVP_CIPHER_CTX_free)>(nullptr, EVP_CIPHER_CTX_free);;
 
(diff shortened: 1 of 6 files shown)


rgacogne (approved):
Looks very good to me! A couple nits, but apart from checking the return value of `RAND_priv_bytes` in a stricter way these are completely cosmetic :)

neilcook:
Thanks Remi, much appreciated!

common/sodcrypto.cc
```diff
@@ -65,20 +64,113 @@ std::string sodDecryptSym(const std::string& msg, const std::string& key, Sodium
...
 
-string newKey()
+std::string sodDecryptSym(const std::string& msg, const std::string& key, SodiumNonce& nonce)
 {
-  return "\"plaintext\"";
-}
+  int len;
+  int plaintext_len;
+  // Each thread gets its own cipher context
+  thread_local EVP_CIPHER_CTX *ctx = nullptr;
```
rgacogne, line 129: Same, I think it should be `static thread_local`.

common/sodcrypto.cc
```diff
@@ -65,20 +64,113 @@ std::string sodDecryptSym(const std::string& msg, const std::string& key, Sodium
...
+
+
+std::string sodEncryptSym(const std::string& msg, const std::string& key, SodiumNonce& nonce)
 {
-  return msg;
+  unsigned char ciphertext[msg.length() + POLY1305_BLOCK_SIZE];
+  int len;
+  int ciphertext_len;
+  // Each thread gets its own cipher context
+  thread_local EVP_CIPHER_CTX *ctx = nullptr;
```
rgacogne, line 87: Just thought about this, don't you want this to be `static thread_local` instead? Otherwise you are not reusing the context.
```suggestion
  static thread_local EVP_CIPHER_CTX *ctx = nullptr;
```

common/sodcrypto.cc
rgacogne, line 129:
```suggestion
  static thread_local EVP_CIPHER_CTX *ctx = nullptr;
```

common/sodcrypto.cc
neilcook: Haha yes good point, I fully intended these to be static but didn’t type it for some reason!

common/sodcrypto.cc
neilcook: 👍

common/sodcrypto.cc
```diff
@@ -65,20 +64,113 @@ std::string sodDecryptSym(const std::string& msg, const std::string& key, Sodium
...
+    throw std::runtime_error("sodEncryptSym: EVP_EncryptUpdate() could not encrypt message");
+  ciphertext_len = len;
+
+  if (1 != EVP_EncryptFinal_ex(ctx, ciphertext + len + POLY1305_BLOCK_SIZE, &len))
+    throw std::runtime_error("sodEncryptSym: EVP_EncryptFinal_ex() could finalize message encryption");;
+  ciphertext_len += len;
+
+  /* Get the tag */
+  if (1 != EVP_CIPHER_CTX_ctrl(ctx, EVP_CTRL_AEAD_GET_TAG, POLY1305_BLOCK_SIZE, ciphertext))
+    throw std::runtime_error("sodEncryptSym: EVP_CIPHER_CTX_ctrl() could not get tag");
```
rgacogne, line 118: One more thing, I'm a bit worried about short-lived threads calling this function, which would result in the context being leaked. I would suggest wrapping the context in a smart pointer instead so that the context is properly released when the thread terminates:

```suggestion
  thread_local auto ctx = std::unique_ptr<EVP_CIPHER_CTX, decltype(&EVP_CIPHER_CTX_free)>(nullptr, EVP_CIPHER_CTX_free);

  if (key.length() != 32) {
    ostringstream err;
    err << "sodEncryptSym: key length is not 32 bytes (actual length=" << key.length() << ")";
    throw std::runtime_error(err.str().c_str());
  }

  if (ctx == nullptr) {
    if (!(ctx = std::unique_ptr<EVP_CIPHER_CTX, decltype(&EVP_CIPHER_CTX_free)>(EVP_CIPHER_CTX_new(), EVP_CIPHER_CTX_free))
      throw std::runtime_error("sodEncryptSym: EVP_CIPHER_CTX_new() could not initialize cipher context");

    /* Initialise the encryption operation. */
    if (1 != EVP_EncryptInit_ex(ctx.get(), EVP_chacha20_poly1305(), NULL, NULL, NULL))
      throw std::runtime_error("sodEncryptSym: EVP_EncryptInit_ex() could not initialize encryption operation");
  }

  if (1 != EVP_EncryptInit_ex(ctx.get(), NULL, NULL, (unsigned char*) key.c_str(), nonce.value))
    throw std::runtime_error("sodEncryptSym: EVP_EncryptInit_ex() could not initialize encryption key and IV");

  if (1 !=
      EVP_EncryptUpdate(ctx.get(), ciphertext + POLY1305_BLOCK_SIZE, &len, (unsigned char*) msg.c_str(), msg.length()))
    throw std::runtime_error("sodEncryptSym: EVP_EncryptUpdate() could not encrypt message");
  ciphertext_len = len;

  if (1 != EVP_EncryptFinal_ex(ctx.get(), ciphertext + len + POLY1305_BLOCK_SIZE, &len))
    throw std::runtime_error("sodEncryptSym: EVP_EncryptFinal_ex() could finalize message encryption");;
  ciphertext_len += len;

  /* Get the tag */
  if (1 != EVP_CIPHER_CTX_ctrl(ctx.get(), EVP_CTRL_AEAD_GET_TAG, POLY1305_BLOCK_SIZE, ciphertext))
    throw std::runtime_error("sodEncryptSym: EVP_CIPHER_CTX_ctrl() could not get tag");
```

neilcook: Yes, this makes a lot of sense.

PR #400, opened 2023-09-29
Docker and Regression Improvements

neilcook:
- Update public Docker base image to debian-bookworm
- Update internal docker base images to latest debian and ubuntu releases 
- Remove Report API from wforce image as it is not used and Report API is going to be deprecated
- Move regression from nose tests (python 2) to pytest (python 3)
- Update deprecated assert functions
- Output junit XML files and create GitHub actions to add to comments in PRs


chbruyand (approved):
LGTM!

PR #410, opened 2023-11-30
Fix build issues with libdrogon

neilcook:
Libdrogon 1.7.4 fails to build on ol-8 and ol-9, however the most recent version 1.9.1 does not build on centos-7 and amazon-2 because that requires openssl-1.1 which is not available on those OSs. This PR updates to 1.9.1 everywhere except when building amazon-2 and centos-7 packages.

Also updates yaml-cpp to the latest version and fixes some missing std:: references in the json11 code.


chbruyand (approved):
LGTM!

Issue #416, opened 2024-01-10
[BUG] License conflicts


P-EB:
**Describe the bug**
Hi,

Some of your code is licensed under GPL-3, other parts (eg [`threadname](https://github.com/PowerDNS/weakforced/blob/master/ext/ext/threadname.cc)) is under GPL-2 (not GPL-2+).

According to [GNU](https://www.gnu.org/licenses/gpl-faq.en.html#AllCompatibility), this is not possible.

Could you relicense any GPL-2 code under "GPL Version 2 or later" ?

This would allow me to package weakforced into Debian.

Thanks !

neilcook:
Sure, this should be possible

neilcook:
I had a look, and only threadname seems to conflict, everything else already GPL-2+, GPL-3 or a compatible license.

P-EB:
<3

P-EB:
@neilcook Thanks!

Do you have plans on a release I could use for Debian? :)

P-EB:
Could you release something like a patch release (ie 2.10.1.1), potentially just having the license fix over 2.10.1 ? Alternatively I can wait a couple months.

neilcook:
Well it would be 2.10.2 not 2.10.1.1. If you can wait a bit that would be great.

P-EB:
Ack

neilcook:
2.10.2 released

neilcook:
JFYI

P-EB:
Didn't reply but saw the release, thanks! <3

Issue #412, opened 2023-12-28
[FEATURE] Split regression so it runs jobs in parallel


jsoref:
**Is your feature request related to a problem? Please describe.**
The regression workflow takes 90mins to run to completion: https://github.com/PowerDNS/weakforced/actions/runs/7342202732 that's too long.

**Describe the solution you'd like**
There should be units of work that can be done in parallel jobs within the workflow.

Whether that's compiling 3 objects in 3 jobs at the beginning and then passing them around via a cache / artifact, or whether that's just running 3 parallel jobs to handle 3 sections of work, ... there has to be something that doesn't depend on previous steps

**Describe alternatives you've considered**
I could wait the time it takes for the workflows to run.

**Additional context**
https://github.com/PowerDNS/weakforced/actions/runs/7342202732 
![image](https://github.com/PowerDNS/weakforced/assets/2119212/772cd340-5d2a-49cf-afaf-6f0828e37215)

neilcook:
Indeed, the regression runs twice, once for clang and once for g++ so that's a prime candidate for running in parallel.

PR #417, opened 2024-01-11
Re-license threadname

neilcook:
The threadname source files were still licensed under GPL-2, so they have been relicensed to GPL-2+ with permission from the author.

Whilst digging around I also discovered that ctpl.h was being included in several places where it had long since stopped being used, so that has also been tidied up in this PR.

The PR also addresses, the long-standing issue of GeoIP tests not working for PRs from forks. Rather than download MaxMind MMDB files, which required repo secrets, I have just created a simple MMDB file as part of the repo that contains all the information needed for the regression tests.

Finally, since I was sick of regression tests taking so long, I've also addressed #412 by splitting the regression tests into two concurrent jobs, which was a fairly simple fix.


neilcook:
Fixes #416

Habbie (approved):

PR #420, opened 2024-04-05
Build minimal docker image

neilcook:
The existing docker image that is built is based on debian and is very heavyweight. This adds a new minimal image based on alpine that is approx 5x smaller, but with less features for casual users.


Habbie (approved):

neilcook:
thx habbie

PR #425, opened 2024-06-04
Use OpenResty Luajit Package with wforce

neilcook:
Regular luajit provided with many distros has memory allocation issues that causes crashes. These issues are fixed in the OpenResty fork of LuaJIT. This PR packages up the OpenResty fork of LuaJIT, and compiles wforce against. The luajit package also include various lua modules that are interesting and useful for wforce such as luasec, elasticsearch and lualdap. Such modules are often only available as packages for later versions of Lua, not Lua 5.1 which is what LuaJIT requires.

This PR also adds support for debian bookworm and drops support for debian buster.

General tidying of the packaging dockerfiles is also part of the PR in order to take better advantage of docker caching/multi-stage builds to improve performance.


builder-support/dockerfiles/Dockerfile.rpmbuild
```diff
@@ -13,31 +13,51 @@ ARG CXX=clang++
...
+##############################################################################
+# Separate build stage for luaenv to speedup builds and improve caching
+
+FROM package-builder-fat-base as luaenv-builder
+
+@IF [ ! -z "$M_all$M_lua" ]
+# These are needed for the luaenv build
+@INCLUDE ../../luaenv/Dockerfile.paths-include
+# Build Lua dist in docker, and later only package it in an RPM spec
+# This is easier to debug that build failures in an RPM build and gets cached in a layer
```
Habbie, line 35:
```suggestion
# This is easier to debug than build failures in an RPM build and gets cached in a layer
```

luaenv/target/bin/luajit
```diff
@@ -0,0 +1 @@
+luajit-2.1.0-beta3
```
Habbie, line 1: the binary this symlink points to - you probably did not mean to check that in?

luaenv/target/bin/luarocks
```diff
@@ -0,0 +1,38 @@
+#!/Users/ncook/GitProjects/weakforced/luaenv/target/bin/luajit
```
Habbie, line 1: nor this

Habbie (changes requested):
halfway through I realised you checked in more than you intended, so I'll continue my review when you cleaned up

neilcook:
> halfway through I realised you checked in more than you intended, so I'll continue my review when you cleaned up

Yes thanks, I didn't run cleanup.sh after a local testing run...

Habbie:
> Yes thanks, I didn't run cleanup.sh after a local testing run...

a bit of `.gitignore` can prevent future accidents

Habbie (approved):
I did not read every line, but the general shape looks good. `build.sh` works on my system, tests pass. Approved (with a double nit from my previous review)

neilcook:
Indeed, will do this before I merge,

Neil

PR #436, opened 2024-08-28
Set 'setKey' and 'webpwd' in debian postinst script only on new installations.

herrwagner:
Change debian postinst scripts to run some shell commands for setting 'setKey' and 'webpwd' parameter only if no weakforce is already installed (or configurations are left). This is necessary because running these commands if weakforce is already runnig they will fail and upgrading exits with failure.
postinst maintainer script is only run _without_ additional arguments beside 'configure' if it's a new installation.
See https://www.debian.org/doc/debian-policy/ch-maintainerscripts.html#summary-of-ways-maintainer-scripts-are-called

### DIFF
--- a/builder-support/debian/wforce-trackalert.postinst
+++ b/builder-support/debian/wforce-trackalert.postinst
@@ -16,11 +16,13 @@ case "$1" in
       adduser --quiet --system --home /var/spool/wforce --shell /bin/false --ingroup wforce --disabled-password --disabled-login --gecos "wforce" wforce
       echo "done"
     fi
-    echo -n "Modifying trackalert.conf to replace password and key..."
-    SETKEY=`echo "makeKey()" | trackalert | grep setKey`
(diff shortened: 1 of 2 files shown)


neilcook:
Would it be possible for you to look into adding these checks also for spec-file based packages?

herrwagner:
> Would it be possible for you to look into adding these checks also for spec-file based packages?

Hi Neil,

as far as i understand the spec file / scriptlets definitions at https://docs.fedoraproject.org/en-US/packaging-guidelines/Scriptlets/ correct - never worked with them or even fedora - the check is already in the weakforce spec files:

https://github.com/PowerDNS/weakforced/blob/e3934ce61997625abafa09d1764fe33e30e8951e/builder-support/specs/wforce.spec#L146

https://github.com/PowerDNS/weakforced/blob/e3934ce61997625abafa09d1764fe33e30e8951e/builder-support/specs/wforce.spec#L178

PR #490, opened 2026-07-07
Add reloadGeoIP2DBs() console command to reload GeoIP2 databases without restart

boxa99:
## Summary

Adds a new console command `reloadGeoIP2DBs()` that reloads all GeoIP2 (MMDB) databases registered with `newGeoIP2DB()` at runtime, without restarting wforce/trackalert — for example after the database files have been updated on disk by geoipupdate.

```
> reloadGeoIP2DBs()
reloadGeoIP2DBs(): Reloaded GeoIP2 DB CityDB
```

## Behaviour

- Each DB is re-opened from its original filename and swapped into `geoip2Map` under the lock. In-flight lookups keep the old DB alive via their `shared_ptr`; it is closed when the last reference is dropped.
- The new DB is opened *before* taking the lock, so lookups are never blocked on disk I/O during a reload.
- If a database file cannot be opened, the existing DB is kept and a per-DB error is reported on the console (and logged), so a partially-failed reload never leaves a DB missing.

## Included

- Implementation in `common/wforce-geoip2.{cc,hh}` and console command registration in `common/common-lua.cc` (registered for wforce and trackalert; a no-op stub in multi-Lua contexts, matching `newGeoIP2DB()`).
- Documentation in the `wforce.1` and `trackalert.1` manpages.
- Regression test (`test_GeoIP.py`) with a second test MMDB fixture (`RegressionGeoIP-v2.mmdb`) covering: a successful reload picks up new data on disk (and reloading the original file restores it), and a failed reload (DB file missing) reports a per-DB error while continuing to serve the old data.


neilcook (approved):
LGTM. Thanks for the feature!

PR #491, opened 2026-07-09
Add Ubuntu LTS build targets (jammy, noble, resolute)

boxa99:
Fixes #468: adds `ubuntu-jammy` (22.04 LTS), `ubuntu-noble` (24.04 LTS) and `ubuntu-resolute` (26.04 LTS) build targets for the package builder, mirroring the existing `debian-trixie` target — each a 12-line `Dockerfile.target.*` file based on the corresponding `ubuntu:` image, registered in the CI matrix and README. `build.sh` auto-discovers them via the existing `Dockerfile.target.*` glob, so no script changes are needed.

As noted in #468, Ubuntu support wasn't completely straightforward — four small fixes surfaced while porting, bundled here because the targets don't build without them:

- **`equivs` installed explicitly** in `Dockerfile.debbuild` / `Dockerfile.target.debuild` — Ubuntu 24.04's `devscripts` no longer pulls it in transitively, so `mk-build-deps` can't create its build-deps meta-package. Harmless on Debian, where `devscripts` still brings `equivs` along.

- **`git` installed explicitly** in `Dockerfile.debbuild` — on Ubuntu 26.04 it no longer arrives transitively either, so the drogon `git clone` in the static-libs stage fails with exit 127. Harmless on the other targets.

- **`apt-get update` before installing built debs** in `Dockerfile.debtest`, matching the existing fix in `builder/helpers/build-debs.sh` — the `dist-base` layer's apt index can be hours stale by the time the install-test stage runs, and version-pinned security-update URLs 404 on the mirror once a point release ships.

- **`wforce-lua-dist` declared as a runtime `Depends`** of `wforce` and `wforce-trackalert`, and its postinst runs `LDCONFIG_NOTRIGGER=y /sbin/ldconfig`. On Ubuntu >= 24.04, `/sbin/ldconfig` invoked from a maintscript defers via `dpkg-trigger`, batching the `ld.so.cache` update to the end of the dpkg run — after `wforce.postinst` has already tried (and failed) to load `libluajit-5.1.so.2`. The env var forces a synchronous update and is ignored on older Debian.

All three targets build and pass the debtest install/verify stage (checked locally on arm64 in addition to the CI runs here).

While touching the README's target list I refreshed it to the current full set of `Dockerfile.target.*` targets (it still referenced the removed `el-7` / `amazon-2`).

### DIFF
--- a/.github/workflows/builder.yml
+++ b/.github/workflows/builder.yml
@@ -21,6 +21,7 @@ jobs:
           - debian-bullseye
           - debian-bookworm
           - debian-trixie
+          - ubuntu-noble
     steps:
       - uses: actions/checkout@v7
         with:
--- a/.github/workflows/builder.yml
+++ b/.github/workflows/builder.yml
@@ -21,7 +21,9 @@ jobs:
           - debian-bullseye
           - debian-bookworm
           - debian-trixie
+          - ubuntu-jammy
           - ubuntu-noble
+          - ubuntu-resolute
     steps:
       - uses: actions/checkout@v7
         with:
--- a/README.md
+++ b/README.md
@@ -58,7 +58,7 @@ $ git clone https://github.com/PowerDNS/weakforced.git
 $ cd weakforced
 $ git submodule init
 $ git submodule update
-$ builder/build.sh debian-bullseye | debian-bookworm | el-7 | el-8  | el-9 | amazon-2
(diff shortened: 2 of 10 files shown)


boxa99:
I think I should also add jammy (22) and resolute (26), to cover all the LTS versions.. Incoming

neilcook:
Thanks for this! I’ll take a look next week when I’m back from vacation

neilcook (approved):

PR #492, opened 2026-07-15
Add TLS support for Redis connections

### LINKED ISSUE #437: [FEATURE] Add SSL support for connection to Redis blacklist
**Is your feature request related to a problem? Please describe.**
These days, all connections need to be SSL, particularly when they are authenticated.  However the wforce server only supports non-ssl connections for the Redis database used for the white/blacklists. 

**Describe the solution you'd like**
Add some configuration flags to allow
* Enable SSL
* Set CA
* Set certificate validation option

**Describe alternatives you've considered**
It is possible to set up a local ssltunnel to forward the connection over SSL, but this is messy

**Additional context**

boxa99:
Closes #437

Adds TLS support for the redis DB used for persistent black/whitelists, via hiredis_ssl.

## New Lua configuration functions

Each has a `whitelist...` equivalent:

- `blacklistRedisTLS(bool)` - enable TLS for the redis connection
- `blacklistRedisTLSCA(ca_file [, ca_path])` - CA to verify the server certificate (defaults to the system CA store)
- `blacklistRedisTLSClientCert(cert_file, key_file)` - client certificate for mutual TLS
- `blacklistRedisTLSVerifyPeer(bool)` - control certificate verification (default true)
- `blacklistRedisTLSServerName(name)` - SNI server name; defaults to the host passed to `blacklistPersistDB()` when it isn't an IP literal

## Build

New `--with-redis-tls` configure option (default `auto`: enabled when `hiredis_ssl >= 1.1.0` is found, warns and builds without otherwise; `yes` makes it a hard requirement; `no` disables). Distros without hiredis_ssl (e.g. ubuntu jammy, debian bullseye) build exactly as before, so no impact on existing package targets.

## Behaviour notes

- The TLS handshake is initiated on every (re)connect, before AUTH, so credentials never travel in plaintext
- The SSL context is rebuilt on reconnect so CA/cert rotation is picked up without a restart
- Enabling TLS in the config of a build without hiredis_ssl fails loudly at startup rather than silently using plaintext
- hiredis verifies the certificate chain but not the hostname; the docs for the new functions note this

## Tests

Adds a regression test that runs a TLS-only redis-server (mutual TLS, reusing the certificates generated by `create-certs.sh`) and verifies blacklist entries persist across a wforce restart over TLS. Skipped automatically when wforce is built without hiredis_ssl or redis-server is absent.

## CI

Bumps the regression image from jammy to noble (following the focal->jammy precedent) so the regression build gets hiredis 1.2.0 with hiredis_ssl and the TLS path is exercised under gcc/clang + ASAN/UBSAN. Also removes two unused pip installs and updates the renamed docker packages.

## Tested

Builds with and without hiredis_ssl; runtime-tested on ubuntu 24.04 against TLS and non-TLS redis servers, including a managed redis service requiring TLS.


boxa99:
@neilcook do you have an idea when you might consider a 3.0.2 (or 3.1.0) release? We have our own patches for this functionality, but we would much rather an upstream fix.

neilcook:
Will be a 3.2 release (we only do even numbered releases for reasons). I’m looking at issuing a release soon, ideally sometime in August but it does depend on availability of folks for testing so may slip until September

boxa99:
Please let me know if I can help with testing

neilcook:
> Please let me know if I can help with testing

Thanks, but this is mostly about testing the builds etc. However if you're interested in the new replication forwarder functionality, you're very welcome to test that!

neilcook:
I added a new test to check that disabling peer verification works. And I refactored the process handling because it was becoming a mess and short-lived process were not always being cleaned up correctly.

neilcook (approved):

